PR #26, opened 2015-09-04
Add option --loop-mode to select loop constraints

jonls:
This removes the option `--no-tfba` from the `fba` command and replaces it with the option `--loop-removal` which takes as parameter either "none", "tfba" or "l1min". The default is set to "none". The "tfba" option applies thermodynamic constraints and the "l1min" performs an L1 flux minimization after the standard FBA.

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -56,6 +56,8 @@ cores on the computer):
 
     $ psamm-model fba --solver threads=4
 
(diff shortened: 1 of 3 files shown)


jonls:
Let me know if you think this proposal is ok @yingzhanguri @keitht547

jonls:
This needs documentation update before merging.

PR #33, opened 2015-09-11
GLPK solver (WIP)

### LINKED ISSUE #9: Add support for open source LP solver
Currently the Cplex and QSoptex solvers are supported. QSoptex is a special case since it is an exact solver, so this leaves Cplex as the only normal solver and also the only solver to support MILP problems. Cplex is proprietary, and although they give out free academic licenses it would be nice to have support for a free, open source solver. GLPK may be the best option.

jonls:
Adds support for the open source (GPL3) LP solver GLPK. Uses the GLPK interface in https://github.com/biosustain/swiglpk. There are still some issues getting `swiglpk` to install correctly on all platforms. The GLPK solver interface currently does not advertise MILP support (in `generic`). The support for MILP problems is in the code but the solver does not seem to respect the integer constraints properly.

Fixes #9.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ addons:
         packages:
             - libgmp-dev
             - libgmp10
+            - glpk
+            - swig
 
 install:
     # Install QSopt_ex library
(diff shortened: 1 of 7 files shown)


jonls:
Blocked on https://github.com/biosustain/swiglpk/issues/8

PR #40, opened 2015-09-28
Tab based export

keitht547:
This implements a new function that can output the information contained within a model as tab based tables. Contains function for reactions, compounds, medium, and limits. 
Needs work for implementing Excel based workbook export

### DIFF
--- a/psamm/commands/excelexport.py
+++ b/psamm/commands/excelexport.py
@@ -1,0 +1,119 @@
+# This file is part of PSAMM.
+#
+# PSAMM is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# PSAMM is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PSAMM.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Copyright 2015  Keith Dufault-Thompson <keitht547@my.uri.edu>
+
+import logging
+
+from ..command import Command, CommandError
+
+try:
+    import xlsxwriter
+except ImportError:
+    xlsxwriter = None
+
+logger = logging.getLogger(__name__)
+
+
+class ExcelExportCommand(Command):
+    """Export the metabolic model as an Excel workbook"""
+
+    @classmethod
+    def init_parser(cls, parser):
+        parser.add_argument(
+            'file', type=str, help='File path for writing the Excel workbook')
+
+    def run(self):
+        model = self._model
+        if xlsxwriter is None:
+            raise CommandError(
+                'Excel export requires the XlsxWriter python module')
+        workbook = xlsxwriter.Workbook(self._args.file)
+        reaction_sheet = workbook.add_worksheet(name='Reactions')
+
+        property_set = set()
+        for j in model.parse_reactions():
+            property_set.update(j.properties)
+        property_list = list(property_set)
+        property_list_sorted = sorted(property_list,
+                                      key=lambda x: (x != 'id',
+                                                     x != 'equation', x))
+        for z, i in enumerate(property_list_sorted):
+            reaction_sheet.write_string(0, z, str(i))
+        for x, i in enumerate(model.parse_reactions()):
+            for y, j in enumerate(property_list_sorted):
+                reaction_sheet.write_string(x+1, y, str(i.properties.get(j)))
+
+        compound_sheet = workbook.add_worksheet(name='Compounds')
+
+        compound_set = set()
+        for compound in model.parse_compounds():
+            compound_set.update(compound.properties)
+
+        compound_list_sorted = sorted(compound_set,
+                                      key=lambda x: (x != 'id',
+                                                     x != 'name', x))
+
+        for z, i in enumerate(compound_list_sorted):
+            compound_sheet.write_string(0, z, str(i))
+        for x, i in enumerate(model.parse_compounds()):
+            for y, j in enumerate(compound_list_sorted):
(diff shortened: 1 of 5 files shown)


jonls:
The build on Travis failed because of some minor issues detected by flake. These should be easy to fix based on the errors reported here: https://travis-ci.org/zhanglab/psamm/jobs/82769704

keitht547:
I'll fix that. It seems like its just formatting things like line length and whitespace.

psamm/commands/table.py
```diff
@@ -0,0 +1,179 @@
+
```
jonls: Copy the copyright note from another file and put it here.

psamm/commands/table.py
```diff
@@ -0,0 +1,179 @@
...
+logger = logging.getLogger(__name__)
+class ExportTableCommand(Command):
+    """Export parts of the model as tab separated tables"""
+
+
+    @classmethod
+    def init_parser(cls, parser):
+        parser.add_argument(
+            'export', metavar='export_type', help='reactions, compounds, medium, limits')
```
jonls: You can add `choices=['reaction', 'compounds', 'medium', 'limits', 'metadata', 'xlsx']` to limit what parameters are accepted.

psamm/commands/table.py
```diff
@@ -0,0 +1,179 @@
...
+        if export_type == 'medium':
+            media_parser(model)
+        if export_type == 'limits':
+            limits_parser(model)
+        if export_type == 'metadata':
+            metadata_parser(model)
+        if export_type == 'xlsx':
+            xlsx_writer(model)
+
+
```
jonls: I would move the following functions into the `ExportTableCommand` class then call them by e.g. `self.reaction_export()`. You don't have to supply the `model` parameter then.

psamm/commands/table.py
```diff
@@ -0,0 +1,179 @@
...
+def reaction_export(model):
+    property_set = set()
+    for j in model.parse_reactions():
+        for i, k in j._properties.iteritems():
```
jonls: Use `j.properties` instead of `j._properties`. Fields starting with an `_` are considered "private" and should generally not be accessed directly.

psamm/commands/table.py
```diff
@@ -0,0 +1,179 @@
...
+            property_set.add(i)
+    property_list = []
+    for i in property_set:
+        property_list.append(i)
```
jonls: Here you can simply do `property_list = list(property_set)` instead of these three lines.

psamm/commands/table.py
```diff
@@ -0,0 +1,179 @@
...
+    for compound in model.parse_compounds():
+        print('\t'.join(str(compound._properties.get(property)) for property in compound_list_sorted))
+
+
+def media_parser(model):
+    for entry in model.parse_media():
+        print('{}\t{}\t{}\t{}'.format('Compound ID', 'Reaction ID', 'Lower Limit', 'Upper Limit'))
+        for media_comp in entry:
+            if media_comp[2] == None:
+                lower = -1000
```
jonls: Instead of `1000` I think this should use the value returned from `model.get_default_flux_limit()`.

psamm/commands/table.py
```diff
@@ -0,0 +1,179 @@
...
+            else:
+                lower = media_comp[2]
+            if media_comp[3] == None:
+                upper = 1000
```
jonls: Also `model.get_default_flux_value()`.

psamm/commands/table.py
jonls: I would call the functions "xxx_writer" or "xxx_export" (pick one for consistency) as you did with `reaction_export` and `xlsx_writer`.

psamm/commands/table.py
```diff
@@ -0,0 +1,179 @@
...
+    print('Default Flux Limits: {}'.format(model.get_default_flux_limit()))
+
+def xlsx_writer(model):
+    workbook = xlsxwriter.Workbook('Model.xlsx')
+    reaction_sheet = workbook.add_worksheet(name='Reactions')
+
+    property_set = set()
+    for j in model.parse_reactions():
+        for i, k in j._properties.iteritems():
+            property_set.add(i)
```
jonls: I think you can do `property_set.update(j.properties)` instead of the two lines above.

psamm/commands/table.py
```diff
@@ -0,0 +1,179 @@
...
+    property_list = []
+    for i in property_set:
+        property_list.append(i)
+    property_list_sorted = sorted(property_list, key=lambda x:(x!='id', x!='equation', x))
+    x = 1
+    z = 0
+    for i in property_list_sorted:
+        reaction_sheet.write_string(0, z, str(i))
+        z = z+1
+    for i in model.parse_reactions():
```
jonls: I would do this:
``` python
for z, i in enumerate(property_list_sorted):
    reaction_sheet.write_string(0, z, str(i))
```

psamm/commands/table.py
```diff
@@ -0,0 +1,179 @@
...
+        y = 0
+        for j in property_list_sorted:
+            reaction_sheet.write_string(x, y, str(i._properties.get(j)))
+            y = y+1
+        x = x+1
```
jonls: Same as above, use `enumerate` whenever possible.

psamm/commands/excelexport.py
```diff
@@ -0,0 +1,110 @@
...
+#
+# PSAMM is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PSAMM.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Copyright 2014-2015  Jon Lund Steffensen <jon_steffensen@uri.edu>
```
jonls: You can change this to `Copyright 2015  Keith ... <...>`

jonls:
I added the tests so the build check pass. I also added a parameter to `excelexport` to specify the file path of the output file. I think this PR is ready for merging now. I will squash some of the commits and merge them if there are no objections.

PR #52, opened 2016-02-01
Let the user specify the compounds to unblock in gapfill

### LINKED ISSUE #43: Allow gapfill of specific compound(s)
Currently the `gapfill` command will first determine all blocked compounds and then try to construct an extended model where all of the compounds are unblocked. This could be extended so that the user can specify a specific subset of the compounds to unblock.

amenge:
Fixes #43

### DIFF
--- a/psamm/commands/gapfill.py
+++ b/psamm/commands/gapfill.py
@@ -19,14 +19,24 @@ from __future__ import unicode_literals
 
 import logging
 
-from ..command import Command, MetabolicMixin, SolverCommandMixin
+from ..command import (Command, MetabolicMixin, SolverCommandMixin,
+                       FilePrefixAppendAction)
+
 from ..gapfill import gapfind, gapfill
+
+from psamm.datasource import reaction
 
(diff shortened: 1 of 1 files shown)


docs/file_format.rst
```diff
@@ -28,6 +29,13 @@ Biomass
 The optional ``biomass`` key specifies the default reaction to use for
 various analyses (e.g. FBA, FVA, etc.)
 
+Extracellular Compartment
+-------------------------
+
+The optional ``extracellular`` key specifies the default string for
+the extracellular comparment on compounds. If this option is not
```
jonls: comparment -> compartment (typo)

docs/file_format.rst
```diff
@@ -28,6 +29,13 @@ Biomass
...
+specified it will be assumed that the extracelluar compartment is called ``e``.
```
jonls: extracelluar -> extracellular

jonls:
@amenge Can you split the last three commits into a separate pull request?

PR #53, opened 2016-02-01
Edit test yaml.py

amenge:


jonls:
I think something went wrong here. The pull request should be based on `develop` not `master`.

PR #65, opened 2016-05-24
native: Check that compound IDs are string types

freestraws:


psamm/datasource/native.py
```diff
@@ -444,8 +455,7 @@ def parse_reaction(reaction_def, context=None):
     """
 
     reaction_id = reaction_def.get('id')
-    if reaction_id is None:
-        raise ParseError('Reaction ID missing')
+    checkid(reaction_id)
```
jonls: The `checkid` function is applied to a reaction ID but the error says "Compound ID ...".

psamm/datasource/native.py
```diff
@@ -307,14 +307,26 @@ def context(self):
         return self._context
 
 
+def checkid(checking):
```
jonls: I would call this `check_id`

psamm/datasource/native.py
```diff
@@ -307,14 +307,26 @@ def context(self):
...
+    """Check whether the ID is valid.
+
+    Does so by checking if type is a string or length is 0."""
+
+    if checking is None:
+        raise ParseError('Compound ID missing')
+    elif type(checking) is not str:
```
jonls: What happens if the ID has unicode characters? You may need to use `isinstance` instead of `type` to get around this ( see https://stackoverflow.com/questions/1549801/differences-between-isinstance-and-type-in-python)

spikeliu:
I continue to develop the new function #86 .

PR #66, opened 2016-05-25
Fix tableexport unicode

freestraws:


jonls:
Looks good! Let's add a test case for this before we merge it.

jonls:
Merging 👍... @spikeliu is working on a test case.

PR #67, opened 2016-05-26
Fix #49: Produce a warning about missing exchange reactions

### LINKED ISSUE #49: Produce a warning about missing exchange reactions
Produce a warning during simulation (FBA, ...?) when a compound defined in the extracellular space does not have an exchange reaction. This warning would be silenced by adding the compound to the medium definition (possibly with bounds set to lower: 0, upper: 0 if no exchange in or out is desired).

spikeliu:


psamm/datasource/native.py
```diff
@@ -283,16 +283,34 @@ def create_metabolic_model(self):
...
+                    extracellular_compounds.add(compound.name)
 
         for compound in sorted(undefined_compounds):
             logger.warning(
                 'The compound {} was not defined in the list'
                 ' of compounds'.format(compound))
 
+        medium_compounds = set()
+        for medium_compound in self.parse_medium():
+            if (medium_compound[0].compartment == extracellular):
```
jonls: No need for parentheses here

PR #72, opened 2016-05-31
Default compartment

spikeliu:
Using extracelluar compartment as the default compartment in medium if user don't provide one.


jonls:
:+1: Please squash into one commit.

PR #82, opened 2016-06-13
Add fastgapfill module and test cases

spikeliu:
Modified and moved the code from `psamm/commands/fastagapfill.py` to a separate module and add test cases for the new module.


psamm/fastgapfill.py
```diff
@@ -0,0 +1,94 @@
...
+# PSAMM is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with PSAMM.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Copyright 2016  Jon Lund Steffensen <jon_steffensen@uri.edu>
+# Copyright 2016  Chao Liu <lcddzyx@gmail.com>
```
jonls: When you copy the copyright notice from another file, also copy the years. I think this should be "2014-2015 Jon ..." and then 2016 for yourself.

psamm/fastgapfill.py
```diff
@@ -0,0 +1,94 @@
...
+
+from six import iteritems
+from .fastcore import fastcore
+
+logger = logging.getLogger(__name__)
+
+
+def create_model_complete(model, db_weight=None, ex_weight=None,
+                          tp_weight=None, penalties=None):
+    """create a complete metabolic model for fastcore algorithm.
```
jonls: Capitalize first letter of sentence. Also, I think this function has to be explained in terms of the fastgapfill procedure described in the paper, i.e. describe that the function creates a MetabolicModel with all reactions added (the reaction database in the model is taken to be the universal database) and also with artificial exchange and transport reactions added.

psamm/tests/test_fastgapfill.py
```diff
@@ -0,0 +1,130 @@
...
+        model_complete, weights = fastgapfill.create_model_complete(
+            self._model,
+            db_weight=1.0,
+            ex_weight=2.0,
+            tp_weight=3.0,
+            penalties=penalties)
+        self.assertEqual(set(model_complete.reactions), reactions)
+        self.assertEqual(weights, w_dict)
+
+        def test_fastgapfill():
```
jonls: This doesn't look right... this test is not run because the indentation makes it a local function defined inside of the other test. I would split these test cases into two classes because the setup is completely different for the two. One requires a NativeModel and the other requires a MetabolicModel.

psamm/tests/test_fastgapfill.py
```diff
@@ -0,0 +1,130 @@
...
+        try:
+            self._solver = generic.Solver()
+        except generic.RequirementsError:
+            self.skipTest('Unable to find an LP solver for tests')
+
+    def tearDown(self):
+        shutil.rmtree(self._model_dir)
+
+    def test_create_model_complete(self):
+        reactions = set([
```
jonls: Maybe call this dict `expected_reactions` and the next dict `expected_weights`.

PR #85, opened 2016-06-20
add randomsparse module and test cases

spikeliu:


psamm/commands/randomsparse.py
```diff
@@ -91,173 +87,20 @@ def run(self):
...
-            len(essential), len(genes)))
+        essential, deleted = randomsparse.random_sparse(
+            strategy, p, reaction, flux_threshold)
 
-        logger.info('Reactions in minimal network: {}'.format(
-            ', '.join(sorted(reactions))))
+        logger.info('Essential entity: {}/{}'.format(
+            len(essential), len(strategy.get_total())))
+        logger.info('Deleted entity: {}/{}'.format(
+            len(deleted), len(strategy.get_total())))
```
jonls: The output should probably be `Essential reactions` or `Essential genes` depending on which strategy was chosen.

psamm/randomsparse.py
```diff
@@ -0,0 +1,207 @@
...
+        self._model = model
+        if reaction_set is None:
+            self._test_set = set(self._model.reactions)
+        else:
+            self._test_set = set(reaction_set)
+        self._reactions = set(self._test_set)
+
+    def get_total(self):
+        return self._reactions
+
```
jonls: Maybe call this function `entities()` since we have adopted this term to refer to reactions or genes. You can even make it a property with

``` python
@property
def entities(self):
    ...
```

psamm/randomsparse.py
```diff
@@ -0,0 +1,207 @@
...
+    def tests(self):
+        while len(self._test_set) > 0:
```
jonls: I'm thinking that `iter_tests` may be a better name than `tests`... to emphasize that this returns an iterator. What do you think?

psamm/randomsparse.py
```diff
@@ -0,0 +1,207 @@
...
+            reaction = random.choice(tuple(self._test_set))
```
jonls: Did we decide against moving the random choice out of this class? I think we discussed an alternative where the list of entities is given in the constructor so that the randomness can (optionally) be done when the strategy is constructed.

psamm/randomsparse.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def delete(self, entity, deleted_reactions):
+        self._reactions -= deleted_reactions
+        self._gene_assoc = self._new_gene_assoc
+
+
+def get_gene_associations(model):
+    """Create gene association for class :class:`.GeneDeletionStrategy`.
+
+    Args:
```
jonls: Add a line describing what it does, e.g. "Return a dict mapping reaction IDs to `boolean.Expression` objects representing the genes that the reaction depends on.".

psamm/randomsparse.py
```diff
@@ -0,0 +1,207 @@
...
+            assoc = boolean.Expression(reaction.genes)
+        else:
+            variables = [boolean.Variable(g) for g in reaction.genes]
+            assoc = boolean.Expression(boolean.And(*variables))
+        yield reaction.id, assoc
+
+
+def get_exchange_reactions(model):
+    """Create reaction ID set for :class:`.GeneDeletionStrategy` if
+    'exchange' is set up as the deletion type.
```
jonls: Describe what the function does in the summary, e.g. "Yield all reaction IDs from model that are exchange reactions". In the next paragraph you can describe that the function is useful for the `GeneDeletionStrategy`. I would also add a note in the `GeneDeletionStrategy` saying that this function can be useful.

psamm/randomsparse.py
```diff
@@ -0,0 +1,207 @@
...
+            yield reaction_id
+
+
+def random_sparse(strategy, prob, obj_reaction, flux_threshold):
+    """Find a random minimal network of model reactions.
+
+    Given a reaction to optimize and a threshold, delete entities randomly
+    until the flux of the reaction to optimize falls under the threshold.
+    Keep deleting until no more entities can be deleted. It works
+    with two strategies: deleting reactions or deleting genes(reactions
```
jonls: Typo: Add space before `(`

psamm/tests/test_randomsparse.py
```diff
@@ -0,0 +1,223 @@
...
+        self._strategy = randomsparse.GeneDeletionStrategy(
+            self._mm, self._assoc)
+
+    def test_method_get_all(self):
+        expected_total = {'gene_1', 'gene_2'}
+        self.assertEqual(set(self._strategy.get_total()), expected_total)
+
+    def test_method_tests_and_delete(self):
+        expected_genes = {'gene_1', 'gene_2'}
+        expected_reaciton_set = {'rxn_2'}
```
jonls: typo: `reaciton` -> `reaction`

psamm/tests/test_randomsparse.py
```diff
@@ -0,0 +1,223 @@
...
+        test_dict = {}
+        first_round = True
+        for gene, deleted_reactions in self._strategy.tests():
```
jonls: I would use `enumerate` here:

``` python
for i, (gene, deleted_reactions) in enumerate(self._strategy.tests()):
    self._strategy.delete(gene, deleted_reactions)
    if i == 0:
        ...
    ...
```

psamm/tests/test_randomsparse.py
```diff
@@ -0,0 +1,223 @@
...
+        self._assoc = {
+            'rxn_1': boolean.Expression('gene_1'),
+            'rxn_2': boolean.Expression('gene_2 or gene_3'),
+            'rxn_5': boolean.Expression('gene_3 and gene_4')
+        }
+
+        self._strategy_react = randomsparse.ReactionDeletionStrategy(self._mm)
+
+        self._strategy_gene = randomsparse.GeneDeletionStrategy(
+            self._mm, self._assoc)
```
jonls: You can move these to the test functions since they are only used in one function each.

psamm/tests/test_randomsparse.py
```diff
@@ -0,0 +1,223 @@
...
+
+        essential, deleted = randomsparse.random_sparse(
+            self._strategy_react,
+            self._prob,
+            self._obj_reaction,
+            flux_threshold=100)
+
+        self.assertTrue(essential in expected_essential)
+        self.assertTrue(deleted in expected_deleted)
+
```
jonls: You can make a more strict test here, I think:

``` python
expected = {'rxn_3', 'rxn_5'}, {'rxn_2', 'rxn_4'}
essential, deleted = randomsparse.random_sparse(
    ...
self.assertTrue(expected == (essential, deleted) or expected == (deleted, essential))
```

This tests that the essential and deleted are in agreement and not containing duplicate elements.

jonls:
:+1:

PR #86, opened 2016-06-20
native: Check that IDs are string types and not empty

spikeliu:


psamm/datasource/native.py
```diff
@@ -307,14 +307,38 @@ def context(self):
         return self._context
 
 
+def _check_id(entity, entity_type):
+    """Check whether the ID is valid.
+
+    First check if the ID is missing, and then check if it is a qualified
+    string type, finally check if the string is empty. For all checks, it
+    would print out corresponding error messages.
```
jonls: "print out" -> "raise a ParseError with the corresponding message"

psamm/datasource/native.py
```diff
@@ -307,14 +307,38 @@ def context(self):
...
+    if entity is None:
+        raise ParseError('{} ID missing'.format(entity_type))
+    elif not isinstance(entity, string_types):
+        msg = '{} ID must be a string, id was {}.'.format(entity_type, entity)
+        if isinstance(entity, bool):
+            msg += (' You may accidentally use "yes", "no", "True" or "False" '
+                    'as the ID. Please check.')
+        raise ParseError(msg)
+    elif len(entity) == 0:
+        raise ParseError('{} ID must be not be empty'.format(entity_type))
```
jonls: "ID must not be empty"

psamm/datasource/native.py
```diff
@@ -307,14 +307,38 @@ def context(self):
...
+            `Compound` or 'Reaction'.
+    """
+
+    if entity is None:
+        raise ParseError('{} ID missing'.format(entity_type))
+    elif not isinstance(entity, string_types):
+        msg = '{} ID must be a string, id was {}.'.format(entity_type, entity)
+        if isinstance(entity, bool):
+            msg += (' You may accidentally use "yes", "no", "True" or "False" '
+                    'as the ID. Please check.')
```
jonls: These are the YAML boolean types: http://yaml.org/type/bool.html

I suggest we say something like: 'You may have accidentally used an ID value that YAML interprets as a boolean, such as "yes", "no", "on", "off", "true" or "false". To use this ID, you have to quote it with single or double quotes.'

PR #88, opened 2016-06-21
Fix various unicode format errors

spikeliu:
Fix various unicode format problems, and add test cases for command with unicode as compound IDs

### DIFF
--- a/psamm/commands/excelexport.py
+++ b/psamm/commands/excelexport.py
@@ -19,6 +19,8 @@ from __future__ import unicode_literals
 
 import logging
 
+from six import text_type
+
 from ..command import Command, CommandError
 
 try:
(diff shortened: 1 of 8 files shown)


psamm/metabolicmodel.py
```diff
@@ -17,6 +17,8 @@
 
 """Representation of metabolic network models."""
 
+from __future__ import unicode_literals
+
```
jonls: This module has a `__repr__` method on `FluxBounds` which needs to be corrected so that it returns `str` type in Python 2.7.

psamm/commands/excelexport.py
```diff
@@ -93,16 +97,16 @@ def run(self):
...
+            media_sheet.write(x+1, 1, text_type(reaction))
+            media_sheet.write(x+1, 2, text_type(lower))
+            media_sheet.write(x+1, 3, text_type(upper))
 
         limits_sheet = workbook.add_worksheet(name='Limits')
 
         limits_sheet.write_string(0, 0, 'Reaction ID')
         limits_sheet.write_string(0, 1, 'Lower Limit')
-        limits_sheet.write_string(0, 2, str('Upper Limit'))
+        limits_sheet.write_string(0, 2, text_type('Upper Limit'))
```
jonls: The `text_type` call is not necessary here and can be removed completely.

psamm/metabolicmodel.py
jonls: :+1:

PR #91, opened 2016-06-23
Genes in model

spikeliu:
It is not correct yet, some tests failed.


jonls:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;HVifAKuv&#x2F;7-split-gene-associations-from-reaction-database"> Split gene associations from reaction database</a>

Issue #74, opened 2016-06-02
In commands, add a separate command error to signal execution errors


jonls:
Currently the `CommandError` exception is available to signal that a command failed. However, this exception causes the argument parser to print out usage information which is only appropriate if the error was caused by incorrect arguments supplied by the user. In other cases, the command may wish to signal an error that does not cause usage information to be printed. Most commands where errors are possible either raise an exception or let an existing exception bubble up. This accomplishes the goal of exiting the command with an error code but does not provide a good error message to the user. Ideally, the stack trace of the exception should be logged (at "debug" level) and a good error message should be logged at "error" level.

jonls:
Fixed in #96

PR #93, opened 2016-06-24
Gapfill epsilon option

jonls:
Add option on `gapfill` command to adjust epsilon parameter (threshold for considering a flux value zero or non-zero). In the second commit, an error message is added when the GapFind or GapFill procedure fails, explaining to the user that the epsilon parameter can be adjusted as a possible solution to the failure. Currently `CommandError` is used to report the error which, in addition, generates a usage message which is not appropriate for this kind of error. This should be changed when #74 is fixed.

### DIFF
--- a/psamm/commands/gapfill.py
+++ b/psamm/commands/gapfill.py
@@ -13,7 +13,7 @@
 # You should have received a copy of the GNU General Public License
 # along with PSAMM.  If not, see <http://www.gnu.org/licenses/>.
 #
-# Copyright 2014-2015  Jon Lund Steffensen <jon_steffensen@uri.edu>
+# Copyright 2014-2016  Jon Lund Steffensen <jon_steffensen@uri.edu>
 
 from __future__ import unicode_literals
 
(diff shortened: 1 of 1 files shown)


psamm/commands/gapfill.py
```diff
@@ -99,3 +110,9 @@ def run(self):
                 print('{}\t{}\t{}'.format(rxnid, 'Reverse', rxt))
         else:
             logger.info('No blocked compounds found')
+
+    def _log_epsilon_and_fail(self, epsilon, exc):
+        msg = ('Finding blocked compounds failed with epsilon set to {}. Try'
+               ' lowering the epsilon value to reduce artifical constraints on'
+               ' the model.'.format(epsilon))
+        self.fail(msg, exc)
```
spikeliu: If this is related to the parameter provided by user, then why this is not a argument error, but a failure, though it is a runtime failure. Or the point is, this is a failure, but user provide what needs to be provided, it's just the epsilon to high or something, so we don't want to show them the argument help information again?

jonls: Yes, that's exactly my rationale for making this a failure. The user has specified the epsilon value correctly but something internally in the command caused a failure.

spikeliu: Then I have no questions about this pull request.

jonls:
@spikeliu Thanks, merging!

PR #94, opened 2016-06-24
command: Add helper methods on Command for error reporting

### LINKED ISSUE #74: In commands, add a separate command error to signal execution errors

jonls:
Fixes #74 by adding the `fail()` method on `Command` which can be used to make the command fail without printing usage information. Optionally, the exception that caused the failure is logged to the debug level.

Todo:
- [ ] Change commands to use `fail` or `error` depending on the type of error (`error`=Error in input parameters, `fail`=Command failure).
- [ ] Rename `error`... perhaps rename to `argument_error`?

### DIFF
--- a/psamm/command.py
+++ b/psamm/command.py
@@ -28,6 +28,7 @@ The :func:`.main` function is the entry point of command line interface.
 from __future__ import division, unicode_literals
 
 import os
+import sys
 import argparse
 import logging
 import abc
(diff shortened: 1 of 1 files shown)


spikeliu:
spikeliu continues this on #96

jonls:
Merged with #96

PR #96, opened 2016-06-27
Command error methods

spikeliu:
I am not sure if providing a string to the function `fail()` is OK, but it seems like that if it is a True value, and it is not a tuple, python would call sys.exc_info() to get the exception info.

Also, I don't know how to test the `argument_error()` and `fail()` using the existing structure in `test_command.py`, I want to construct a Command instance, but it's an abstract base class. To test these two functions, I don't want to run any actual command either.

### DIFF
--- a/psamm/command.py
+++ b/psamm/command.py
@@ -28,6 +28,7 @@ The :func:`.main` function is the entry point of command line interface.
 from __future__ import division, unicode_literals
 
 import os
+import sys
 import argparse
 import logging
 import abc
@@ -90,6 +91,17 @@ class Command(object):
     @abc.abstractmethod
     def run(self):
         """Execute command"""
+
+    def argument_error(self, msg):
+        """Raise error indicating error parsing an argument."""
+        raise CommandError(msg)
+
+    def fail(self, msg, exc=None):
+        """Exit command as a result of a failure."""
+        logger.error(msg)
+        if exc is not None:
+            logger.debug('Command failure caused by exception!', exc_info=exc)
+        sys.exit(1)
 
 
 class MetabolicMixin(object):
--- a/psamm/commands/fastgapfill.py
+++ b/psamm/commands/fastgapfill.py
@@ -20,7 +20,7 @@ from __future__ import unicode_literals
 import argparse
 import logging
 
-from ..command import Command, MetabolicMixin, SolverCommandMixin, CommandError
+from ..command import Command, MetabolicMixin, SolverCommandMixin
 from .. import fastcore, fluxanalysis
 from ..fastgapfill import create_extended_model, fastgapfill
 
@@ -105,12 +105,12 @@ class FastGapFillCommand(MetabolicMixin, SolverCommandMixin, Command):
         else:
             maximized_reaction = self._model.get_biomass_reaction()
             if maximized_reaction is None:
-                raise CommandError('The maximized reaction was not specified')
+                self.argument_error(
+                    'The maximized reaction was not specified')
 
         if not self._mm.has_reaction(maximized_reaction):
-            raise CommandError(
-                'The biomass reaction is not a valid model'
-                ' reaction: {}'.format(maximized_reaction))
+            self.fail('The biomass reaction is not a valid model'
+                      ' reaction: {}'.format(maximized_reaction))
 
         logger.info('Flux balance on induced model maximizing {}'.format(
             maximized_reaction))
--- a/psamm/commands/fba.py
+++ b/psamm/commands/fba.py
@@ -20,7 +20,7 @@ from __future__ import unicode_literals
 import time
 import logging
 
-from ..command import SolverCommandMixin, MetabolicMixin, Command, CommandError
+from ..command import SolverCommandMixin, MetabolicMixin, Command
 from .. import fluxanalysis
 
 logger = logging.getLogger(__name__)
@@ -65,11 +65,11 @@ class FluxBalanceCommand(MetabolicMixin, SolverCommandMixin, Command):
         else:
             reaction = self._model.get_biomass_reaction()
             if reaction is None:
-                raise CommandError('The biomass reaction was not specified')
+                self.argument_error('The biomass reaction was not specified')
 
         if not self._mm.has_reaction(reaction):
-            raise CommandError('Specified reaction is not in model: {}'.format(
-                reaction))
+            self.fail(
+                'Specified reaction is not in model: {}'.format(reaction))
 
         logger.info('Using {} as objective'.format(reaction))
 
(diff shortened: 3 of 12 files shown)


psamm/commands/fastgapfill.py
```diff
@@ -105,12 +105,11 @@ def reaction_key(r):
         else:
             maximized_reaction = self._model.get_biomass_reaction()
             if maximized_reaction is None:
-                raise CommandError('The maximized reaction was not specified')
+                self.fail('The maximized reaction was not specified')
```
jonls: I think this should be an argument error.

psamm/command.py
```diff
@@ -91,6 +92,16 @@ def init_parser(cls, parser):
     def run(self):
         """Execute command"""
 
+    def argument_error(self, msg):
+        """Raise error indicating error parsing an argument."""
+        raise CommandError(msg)
+
+    def fail(self, exc=None):
```
jonls: The `exc` parameter would be in instance of `Exception` indicating the exception that caused the command to fail. You can expand the docstring with an explanation of this.

psamm/commands/fastgapfill.py
```diff
@@ -105,12 +105,11 @@ def reaction_key(r):
...
 
         if not self._mm.has_reaction(maximized_reaction):
-            raise CommandError(
-                'The biomass reaction is not a valid model'
-                ' reaction: {}'.format(maximized_reaction))
+            self.fail('The biomass reaction is not a valid model'
+                      ' reaction: {}'.format(maximized_reaction))
```
jonls: Same as above... ?

psamm/commands/fba.py
```diff
@@ -65,11 +65,11 @@ def run(self):
         else:
             reaction = self._model.get_biomass_reaction()
             if reaction is None:
-                raise CommandError('The biomass reaction was not specified')
+                self.fail('The biomass reaction was not specified')
```
jonls: See comment above

psamm/commands/fluxcoupling.py
```diff
@@ -43,7 +43,7 @@ def run(self):
 
         max_reaction = self._model.get_biomass_reaction()
         if max_reaction is None:
-            raise CommandError('The biomass reaction was not specified')
+            self.fail('The biomass reaction was not specified')
```
jonls: See comment above

psamm/commands/fva.py
```diff
@@ -60,11 +60,11 @@ def run(self):
         else:
             reaction = self._model.get_biomass_reaction()
             if reaction is None:
-                raise CommandError('The biomass reaction was not specified')
+                self.fail('The biomass reaction was not specified')
```
jonls: See comment above

psamm/commands/genedelete.py
```diff
@@ -48,7 +48,7 @@ def run(self):
         else:
             obj_reaction = self._model.get_biomass_reaction()
             if obj_reaction is None:
-                raise CommandError('The biomass reaction was not specified')
+                self.fail('The biomass reaction was not specified')
```
jonls: See comment above

psamm/commands/randomsparse.py
```diff
@@ -61,11 +61,11 @@ def run(self):
         else:
             reaction = self._model.get_biomass_reaction()
             if reaction is None:
-                raise CommandError('The biomass reaction was not specified')
+                self.fail('The biomass reaction was not specified')
```
jonls: See above

psamm/commands/robustness.py
```diff
@@ -78,20 +78,20 @@ def run(self):
         else:
             reaction = self._model.get_biomass_reaction()
             if reaction is None:
-                raise CommandError('The biomass reaction was not specified')
+                self.fail('The biomass reaction was not specified')
```
jonls: See above

psamm/commands/fastgapfill.py
spikeliu: I think this one should be kept as `fail()`, as above, it has already called `argument_error()`. Something like, there is no problem to get the biomass reaction user specified either in the model or in the command line, but it is not a valid one, so the program fails.

psamm/commands/fluxcoupling.py
spikeliu: this command is not trying to read the reaction from command line, but trying to get it from the model method, so it fails, it should just be a run time fail, right?

psamm/commands/robustness.py
spikeliu: There are two places that I think should keep the way it is. See the comments I made above. If you agree, I would git push a new version.

psamm/command.py
spikeliu: But now I provide a string to it, but it should still be OK, right? Because it should be constructed by `error_info()` function?

psamm/commands/fluxcoupling.py
jonls: Yes, I think you are right.

psamm/command.py
jonls: No `exc` should always be an `Exception`. Otherwise the log message doesn't make any sense. You can add another parameter with a string error message that you log to `logger.error` before logging the `exc`.

psamm/commands/fastgapfill.py
jonls: I agree, we can keep the ones that check whether the reaction is a valid model reaction as `fail`.

jonls:
I think the way you test these methods is fine. You could do something more elaborate where you create new mock `Command`s that fail but I don't think it's worth it.

psamm/command.py
spikeliu: But I did see the descriptions in the official website saying that when exc is not False value, if exc is a tuple, it is used; otherwise `sys.exc_info()` is called to get the exception information. Do I misunderstand this part? https://docs.python.org/2/library/logging.html

psamm/command.py
```diff
@@ -91,6 +92,18 @@ def init_parser(cls, parser):
...
+        raise CommandError(msg)
+
+    def fail(self, msg, exc=None):
+        """Exit command as a result of a failure."""
+        if exc is not None:
+            logger.debug(msg, exc_info=exc)
+        else:
+            logger.debug(msg)
+        sys.exit(1)
+
```
jonls: I think you misunderstood this... the `msg` should be logged at the `error` level otherwise a normal user will not know what happened. In other words, `logger.error` should be used to log `msg`. In addition, if `exc` is provided, the exception should be logged with `logger.debug`. This is to provide additional log messages about the exception traceback so that a developer can trace the source of the failure.

jonls:
Great 👍  Can you squash my commit and your first commit? Also, your last commit message should be sentence cased.

PR #97, opened 2016-06-28
sbml: writer now uses NativeModel instead of MetabolicModel

spikeliu:
It's hard to take a look at the differences in the output, as they are grouped in just one line. I think there are something could be changed: 1) reaction id generated by the writer() to each reaction; 2) the flux limit for each reaction, as in NativeModel, the direction of each reaction could be forward, reverse or both, I assume using MetabolicModel would only generates forward and both; 3) I use 1000.0 and 0.0 and the default_flux_limit and default_flux_limit_blocked, so it may not be integer 1000 and 0 in some cases.

The difficult part lies in the flux limit part.

### DIFF
--- a/psamm/datasource/sbml.py
+++ b/psamm/datasource/sbml.py
@@ -648,7 +648,7 @@ class SBMLWriter(object):
         self._namespace = SBML_NS_L3_V1_CORE
         self._sbml_tag = partial(_tag, namespace=self._namespace)
 
-    def write_model(self, file, model, reactions, compounds):
+    def write_model(self, file, model):
         """Write a given model to file"""
 
         ET.register_namespace('mathml', MATHML_NS)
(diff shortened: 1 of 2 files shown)


psamm/datasource/sbml.py
```diff
@@ -31,6 +31,10 @@
 from ..reaction import Reaction, Compound, Direction
 
 
+# Default flux limit value
+_DEFAULT_FLUX_LIMIT = 1000.0
```
jonls: You can get this value from `NativeModel` with `get_default_flux_limit()`.

psamm/datasource/sbml.py
```diff
@@ -31,6 +31,10 @@
...
+_DEFAULT_FLUX_LIMIT_BLOCKED = 0.0
```
jonls: I think you can assume this to be always zero. No need for a constant.

psamm/datasource/sbml.py
```diff
@@ -713,20 +720,41 @@ def write_model(self, file, model, reactions, compounds):
...
+                lower = flux_limit[2]
+            # Deal with upper limit
+            if flux_limit[1] is None:
+                upper = (
+                    _DEFAULT_FLUX_LIMIT_BLOCKED
+                    if reaction_equations[eq_id].direction == Direction.Reverse
+                    else _DEFAULT_FLUX_LIMIT)
+            else:
+                upper = flux_limit[1]
+            flux_limits[eq_id] = {'lower': lower, 'upper': upper}
```
jonls: I would probably just make this a 2-tuple.

jonls:
Did you test to see that this generates SBML files that are identical to the files generated before? Try to take a couple of models from the model collection and convert them back to SBML files using PSAMM before/after this patch. After generating the SBML files you can apply `sed -e 's|>|>'$'\\\n''|g'` to make it easier to `diff` the before/after files.

I suspect that you are not taking care of generating the exchange reactions in the SBML files with this patch, since those reactions are represented as a medium definition in `NativeModel` but are represented as a regular reaction in `MetabolicModel`.

spikeliu:
It is hard to check as the IDs and orders of reactions and compounds are different, if you say there is a potential, then I would go back and check how NativeModel generate the Metabolic model. I think this is also related to how we import the sbml model to yaml format, if we don't set no-medium, then it could happen, right?
I also remember that the old function gets gene-reaction association from nativemodel.parse_reactions(),  but since they are artifical reactions in medium, so they would never have gene associations, right?

spikeliu:
One test related to NativeModel.get_default_flux_limit() is not working, as it could return None for now, I think after change its behavior, then everything would be fine.
I checked old and new sbml files, they are of the same number of lines now and they are very similar. That's the only thing I can tell.

jonls:
@spikeliu What are we currently stuck on with this PR? Did you try to set up a test with the `sed` and `diff -u` that I proposed above? You may have to apply a sort when writing out the model to get consistent SBML files.

spikeliu:
This is a long time ago, I did remember I did some tests, the line number of two files are the same. I would try to figure pick what I left before.

jonls:
Merged in #113.

PR #100, opened 2016-06-30
Release 0.22

spikeliu:
Release Candidate for PSAMM v0.22


NEWS.md
```diff
@@ -1,3 +1,21 @@
...
+- When parsing reaction and compound entities from the YAML files, add
+  abilities to check if the ID strings are legitimate to be read in.
+- Improve how PSAMM handle different errors when called from command line, it
+  would prompt proper error messages for different error types.
+- If a model doesn't have a default setup for default flux limit,
+  `NativeModel.get_default_flux_limit()` would return 1000 instead of `None`.
+- Fix a bug where calling get_num_binary() on the Cplex instance can cause a
+  segment fault in some rare cases.
+- Fix various unicode handling problems in YAML files and some typos in the
+  docstrings.
```
jonls: Let's prioritize these and exclude things that are not user visible. We can argue that API is user visible but I would put them at the bottom of the list and mark them as such. I think the most important change is handling unicode properly in commands so that should go first.

- I don't think checking IDs in YAML files is important enough for this file. Maybe we can explain it differently: "When parsing reaction and compound entities from YAML files, produce better error messages when IDs are invalid".
- I don't think the error handling is important enough.
- The change to `get_default_flux_limit()` is a minor API change. The default value was always 1000, it is just defined in a different place now. I don't think it is important enough.
- For the Cplex bug fix, I would not mention the technical details. Something like: "Work around a bug in Cplex that in rare causes a segmentation fault when a linear programming problem is solved repeatedly".
- Fixing typos are generally not important enough for the NEWS file.
- At the end, I would put the notes on the API additions. Something like "API: Add `fastgapfill` module which allows access to fun the fastGapFill algorithm" and similar for `randomsparse`.

NEWS.md
```diff
@@ -1,3 +1,16 @@
+v0.22 (2016-06-30)
+------------------
+
+- Better unicode handling in commands.
+- When parsing reaction and compound entities from the YAML files, produce
+  better error messages when IDs are invalid.
+- Work around a bug in Cplex that in rare causes a segmentation fault when a
```
jonls: "in rare causes" -> "in rare cases results in a segmentation fault..."

NEWS.md
```diff
@@ -1,3 +1,16 @@
...
+  linear programming problem is solved repeatedly.
+- Add `fastgapfill` module which allows API access to run fastgapfill
+    algorithm.
+- Add `randomsparse` module which allows API access to generate a random
+    minimum model which satisfies the flux threshold of the objective reaction.
```
jonls: Maybe prefix these two with API? "API: Add `fastgapfill` module which allow access to run the fastGapFill algorithm."

jonls:
I have made the final changes to the NEWS file and merged this 👍

PR #102, opened 2016-07-05
New dictdatabase

spikeliu:
Extend `DictDatabase` to hold all info from yaml files.
Add `create_database()` method in NativeModel.

### DIFF
--- a/psamm/database.py
+++ b/psamm/database.py
@@ -179,7 +179,7 @@ class DictDatabase(MetabolicDatabase):
         # reaction map from reaction ID to ReactionEntry
         self._reaction_entries = defaultdict(dict)
         # original attributes in DictDatabase, consider to be moved
-        self._reactions = defaultdict(dict)
+        self._reaction_values = defaultdict(dict)
         self._compound_reactions = defaultdict(set)
         self._reversible = set()
 
@@ -196,8 +196,12 @@ class DictDatabase(MetabolicDatabase):
         return self._default_flux_limit
 
     @property
+    def reaction_entries(self):
+        return iter(self._reaction_entries)
+
+    @property
     def reactions(self):
-        return iter(self._reactions)
+        return iter(self._reaction_values)
 
     @property
     def compound_entries(self):
@@ -224,15 +228,15 @@ class DictDatabase(MetabolicDatabase):
         return compound_id in self._compound_entries
 
     def has_reaction(self, reaction_id):
-        return reaction_id in self._reaction_entries
+        return reaction_id in self._reaction_values
 
     def is_reversible(self, reaction_id):
         return reaction_id in self._reversible
 
     def get_reaction_values(self, reaction_id):
-        if reaction_id not in self._reactions:
+        if reaction_id not in self._reaction_values:
             raise ValueError('Unknown reaction: {}'.format(repr(reaction_id)))
-        return iteritems(self._reactions[reaction_id])
+        return iteritems(self._reaction_values[reaction_id])
 
     def get_compound_reactions(self, compound):
         return iter(self._compound_reactions[compound])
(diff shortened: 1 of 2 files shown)


psamm/database.py
```diff
@@ -164,28 +164,67 @@ class DictDatabase(MetabolicDatabase):
...
+        # _medium is a map from reaction_id to (Compound, upper, lower)
+        self._medium = {}
+        # reaction IDs in the model
+        self._model = set()
+        # compound map from compound ID to CompoundEntry
+        self._compound_entries = defaultdict(dict)
+        # reaction map from reaction ID to ReactionEntry
+        self._reaction_entries = defaultdict(dict)
+        # original attributes in DictDatabase, consider to be moved
         self._reactions = defaultdict(dict)
```
spikeliu: How about change this attribute name to _reaction_values in the future now matter where it would be? Also for the related code using this attribute.

psamm/database.py
```diff
@@ -164,28 +164,67 @@ class DictDatabase(MetabolicDatabase):
...
+    def extracelluar_compartment(self):
+        return self._extracellular_compartment
+
+    @property
+    def defalut_flux_limit(self):
+        return self._default_flux_limit
+
+    @property
     def reactions(self):
         return iter(self._reactions)
```
spikeliu: I think this part should be change to return _reaction_entries. This depends on how we deal with the exchange reactions in the medium.yaml in the future.

psamm/database.py
```diff
@@ -164,28 +164,67 @@ class DictDatabase(MetabolicDatabase):
...
+        # reaction IDs in the model
+        self._model = set()
+        # compound map from compound ID to CompoundEntry
+        self._compound_entries = defaultdict(dict)
+        # reaction map from reaction ID to ReactionEntry
+        self._reaction_entries = defaultdict(dict)
+        # original attributes in DictDatabase, consider to be moved
         self._reactions = defaultdict(dict)
         self._compound_reactions = defaultdict(set)
         self._reversible = set()
```
spikeliu: For these 3 original attributes, do you think we keep them here or move them in MetabolicModel.

psamm/database.py
```diff
@@ -164,28 +164,67 @@ class DictDatabase(MetabolicDatabase):
```
spikeliu: For this class, its parent class has some method like get_reaction(), may also need to change depending on what we want to keep in the database.

psamm/database.py
```diff
@@ -164,28 +164,67 @@ class DictDatabase(MetabolicDatabase):
...
             if compound.compartment not in compartment_set:
                 compartment_set.add(compound.compartment)
                 yield compound.compartment
 
+    def has_compound(self, compound_id):
+        return compound_id in self._compound_entries
+
     def has_reaction(self, reaction_id):
-        return reaction_id in self._reactions
+        return reaction_id in self._reaction_entries
```
spikeliu: This change would cause some tests fail, as it doesn't have exchange reaction in _reaction_entries

psamm/database.py
```diff
@@ -195,8 +234,32 @@ def get_reaction_values(self, reaction_id):
...
+        return iter(self._compound_reactions[compound])
+
+    def set_model_name(self, model_name):
+        self._model_name = model_name
+
+    def set_extracellular_compartment(self, extracelluar_compartment):
+        self._extracellular_compartment = extracelluar_compartment
+
+    def set_default_flux_limit(self, flux_limit):
+        self._default_flux_limit = flux_limit
```
jonls: Instead of these three methods, use the properties above to make these settable. See the documentation on `@property` on how to do this.

jonls:
What is the conceptual distinction between DictDatabase and MetabolicModel with these changes? Previously, the database held only general reaction information whereas the model held information related to a specific model i.e. a specific collection of reactions with flux limits. The flux limits are now moved to the DictDatabase and the DictDatabase also seems to contain a model subset. Is there a justification for having a separate MetabolicModel class or is the proposal to completely merge the functionality of MetabolicModel into DictDatabase?

spikeliu:
For now, I just make DictDatabase hold all info from yaml files, and also, if we define it as a database, then it could also have the relationship between different concept or objects, like compound_reactions holds the relation between compound and reactions, and the original `reactions`(to call it reaction_values would be more clear) holds the relation between reaction and its stoichiometric values for each compound. I think `MetabolicModel` could continue to access these data. I'm not sure to move them out to MeatabolicModel or just keep them here, it may depend on what kind of new feature we want to give MetabolicModel like those flexible ways to access reactions.

Issue #71, opened 2016-05-31
Extend compatibility with older Cplex versions by only using set_offset when available


jonls:
Currently the `set_offset()` method is used with the Cplex solver when setting the LP problem objective. Using the `set_offset()` method is not strictly necessary to get the correct solution but does allow Cplex to report the correct objective value when a constant term is included in the objective. However, `set_offset()` is only available starting from 12.6.2 which means that PSAMM fails to work with older versions of Cplex. We can extend compatibility to older versions of Cplex by only using `set_offset()` when available.

jonls:
Fixed in #104

PR #104, opened 2016-07-13
cplex: add version check before calling set_offset() method

spikeliu:
#71

### DIFF
--- a/psamm/lpsolver/cplex.py
+++ b/psamm/lpsolver/cplex.py
@@ -246,7 +246,8 @@ class Problem(BaseProblem):
         if len(quad) > 0:
             self._cp.objective.set_quadratic_coefficients(quad)
 
-        self._cp.objective.set_offset(expression.offset)
+        if hasattr(self._cp.objective, 'set_offset'):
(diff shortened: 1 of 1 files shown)


psamm/lpsolver/cplex.py
```diff
@@ -246,7 +246,8 @@ def set_objective(self, expression):
         if len(quad) > 0:
             self._cp.objective.set_quadratic_coefficients(quad)
 
-        self._cp.objective.set_offset(expression.offset)
+        if self._cp.get_version() >= '12.6.2.0':
+            self._cp.objective.set_offset(expression.offset)
```
jonls: I don't think this is a good way of comparing versions. For example,

``` python
>>> '9.8' >= '12.6.2.0'
True
```

Instead, I would just check that the method exists:

``` python
if hasattr(self._cp.objective, 'set_offset'):
    ...
```

spikeliu: Yeah, I didn't think of that. All in my mind is version check...

jonls:
:+1:

PR #108, opened 2016-07-28
Improved metabolic interface

spikeliu:


jonls:
The Travis check failed. @spikeliu Do you want to fix this or should we hand this off to @freestraws ?

psamm/metabolicmodel.py
```diff
@@ -145,12 +145,34 @@ def __getitem__(self, key):
...
 
     def __iter__(self):
-        return self._model.reactions
+        return iter(self._model.reactions)
 
     def __len__(self):
         return sum(1 for _ in self._model.reactions)
 
 
+class _ListWithKey(list):
```
jonls: Subclassing `list` seems to be discouraged. Better to implement the `collections.Sequence` interface. This also avoids the awkwardness of overriding `__setitem__` to raise a `RuntimeError`.

psamm/metabolicmodel.py
```diff
@@ -174,11 +196,11 @@ def database(self):
 
     @property
     def reactions(self):
-        return iter(self._reaction_set)
+        return self._reaction_set
 
     @property
     def compounds(self):
-        return iter(self._compound_set)
+        return self._compound_set
```
jonls: `compounds` is changed to return the private `set` which is mutable. This allows the caller to accidentally modify the set of compounds which is bad. Instead an immutable view on the set should be returned.

psamm/metabolicmodel.py
```diff
@@ -174,11 +196,11 @@ def database(self):
 
     @property
     def reactions(self):
-        return iter(self._reaction_set)
+        return self._reaction_set
```
jonls: The returned value is not immutable. Although the `__setitem__` raises an exception the `append` method (and others) is still available so this allows the caller to modify the list of reactions without going through `add_reaction()` which would violate the class invariant.

jonls:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;mOttN2nn&#x2F;24-refine-metabolicmodel-interface-to-deal-with-both-index-and-key-search-for-reactions-also-retain-the-order-of-reactions-and-comp"> Refine MetabolicModel interface to deal with both index and key search for reactions. Also retain the order of reactions and compounds when loading it into the DictDatabase.</a>

PR #109, opened 2016-08-01
Extra col tableexport

spikeliu:
add an extra column for `tableexport` module to indicate whether a certain reaction or compound is in the model definition.

### DIFF
--- a/psamm/commands/tableexport.py
+++ b/psamm/commands/tableexport.py
@@ -62,16 +62,31 @@ class ExportTableCommand(Command):
 
     def _reaction_export(self):
         property_set = set()
-        for j in self._model.parse_reactions():
-            property_set.update(j.properties)
+        for reaction in self._model.parse_reactions():
+            property_set.update(reaction.properties)
 
         property_list_sorted = sorted(
             property_set, key=lambda x: (x != 'id', x != 'equation', x))
 
-        print('\t'.join([text_type(x) for x in property_list_sorted]))
-        for i in self._model.parse_reactions():
-            print('\t'.join(text_type(i.properties.get(property))
-                            for property in property_list_sorted))
+        print('\t'.join(
+            [text_type(x) for x in property_list_sorted] + ['in_model']))
+
+        if self._model.has_model_definition():
+            model_reactions = set(self._model.parse_model())
+            for reaction in self._model.parse_reactions():
+                line_content = [text_type(reaction.properties.get(property))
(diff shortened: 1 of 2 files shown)


psamm/commands/tableexport.py
```diff
@@ -62,16 +62,31 @@ def run(self):
...
+        if self._model.has_model_definition():
+            model_reactions = set(self._model.parse_model())
+            for reaction in self._model.parse_reactions():
+                line_content = [text_type(reaction.properties.get(property))
+                                for property in property_list_sorted]
+                if reaction.id in model_reactions:
+                    line_content.append('True')
+                else:
+                    line_content.append('False')
+                print('\t'.join(line_content))
```
jonls: Wouldn't it be possible to remove the second loop below by changing the inner `if` statement to

``` python
if not self._model.has_model_definition() or reaction_id in model_reactions:
    line_content.append(text_type(True))
else:
...
```

psamm/commands/tableexport.py
```diff
@@ -81,10 +96,25 @@ def _compound_export(self):
...
+            metabolic_model = self._model.create_metabolic_model()
+            model_compounds = set(x.name for x in metabolic_model.compounds)
+            for compound in self._model.parse_compounds():
+                line_content = [text_type(compound.properties.get(property))
+                                for property in compound_list_sorted]
+                if compound.id in model_compounds:
+                    line_content.append('True')
+                else:
+                    line_content.append('False')
+                print('\t'.join(line_content))
```
jonls: Same as above?

psamm/commands/tableexport.py
spikeliu: The reason I wrote in this way is that I don't want to add extra operations on judging `self._model.has_model_definition()` multiple times. In your way, we would do the same check over and over again.

jonls: The check should be very cheap. My guess is you'll have a hard time measuring any difference in performance but maintainability goes down by having two copies of the loop. `tableexport` is not performance critical anyway.

spikeliu: Yeah, but this is easy to modify as there is no need to call `model.create_metabolic_model()`, but the next one for compounds would be first check if there is a model definition and then create the model, and then it would still not be so elegant.

jonls: I'm not sure why you are creating a `MetabolicModel`. Why not get the information from `NativeModel`?

Issue #89, opened 2016-06-21
Allow reactions with fixed flux in the limits definition


jonls:
Currently to fix the flux of a reaction to a specific value it is necessary to specify a lower and upper bound of that specific value. Instead, we can allow a key `fixed` as a shorthand for setting both lower and upper bound to the same value. When this is implemented, psamm-import should be changed to use this format.

jonls:
Make sure to generate an error if both `fixed` and one of `lower` or `upper` are set at the same time.

PR #110, opened 2016-08-04
native: Made it so the 'fixed' key is recognized in limits

freestraws:
#89


jonls:
See my comment about generating error messages in #89. If you detect that both `fixed` and `lower` is set or both `fixed` and `upper` is set, I think we should generate an error.

jonls:
Do you have a small test model that shows that this works? Do you also have a test that shows that the error message is generated?

Issue #80, opened 2016-06-07
Make NativeModel constructor take string, dict or file object


spikeliu:
NativeModel currently only receives exact file name as the parameter. For creating test cases, it's not flexible enough. So, in addition, it should be able to take a string or file object to parse it, or take a dict object and use it directly.

PR #112, opened 2016-08-11
native: NativeModel can take long strings, dicts, or open files as model

freestraws:
https://github.com/zhanglab/psamm/issues/80

### DIFF
--- a/psamm/datasource/native.py
+++ b/psamm/datasource/native.py
@@ -174,28 +174,45 @@ class NativeModel(object):
     directory subtree that specifies part of the model.
     """
 
-    def __init__(self, path):
-        """Create a model from the specified model file or directory"""
+    def __init__(self, model_from, context=None):
+        """Create a model from the specified content.
 
-        if os.path.isfile(path):
-            self._context = FilePathContext(path)
-            with open(self._context.filepath, 'r') as f:
-                self._model = yaml_load(f)
+        Model can be a string, open file, or dictionary.
+        """
+        if isinstance(model_from, string_types):
+            self._model = yaml_load(model_from)
+            self._context = context
+        elif isinstance(model_from, dict):
+            self._context = context
+            self._model = model_from
+        elif hasattr(model_from, 'read') and callable(model_from.read):
+            self._context = context
+            self._model = yaml_load(model_from)
         else:
+            raise ValueError("Model is of an invalid types")
+
+    @classmethod
+    def load_model_from_path(cls, path):
+        """Create a model from specified path."""
+        context = FilePathContext(path)
+        try:
+            with open(context.filepath, 'r') as f:
+                return NativeModel(f, context)
+        except IOError:
             # Try to open the default file
             for filename in DEFAULT_MODEL:
                 try:
-                    self._context = FilePathContext(
+                    context = FilePathContext(
                         os.path.join(path, filename))
-                    with open(self._context.filepath, 'r') as f:
-                        self._model = yaml_load(f)
-                        break
+                    with open(context.filepath, 'r') as f:
+                        return NativeModel(f, context)
                 except:
-                    logger.debug('Failed to load model file', exc_info=True)
-            else:
-                # No model could be loaded
-                raise ParseError('No model file could be found ({})'.format(
-                    ', '.join(DEFAULT_MODEL)))
+                    logger.debug('Failed to load model file',
+                                 exc_info=True)
+
+        # No model could be loaded
+        raise ParseError('No model file could be found ({})'.format(
+            ', '.join(DEFAULT_MODEL)))
 
     def get_name(self):
         """Return the name specified by the model"""
--- a/psamm/tests/test_yaml.py
+++ b/psamm/tests/test_yaml.py
@@ -21,7 +21,7 @@ import shutil
 import tempfile
 import unittest
 
-from psamm.datasource import native
+from psamm.datasource import native, context
 from psamm.reaction import Reaction, Compound, Direction
 
 from six import StringIO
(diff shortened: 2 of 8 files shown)


psamm/datasource/native.py
```diff
@@ -174,28 +174,38 @@ class NativeModel(object):
     directory subtree that specifies part of the model.
     """
 
-    def __init__(self, path):
+    def __init__(self, model_from):
         """Create a model from the specified model file or directory"""
```
jonls: This docstring needs to be updated.

psamm/datasource/native.py
```diff
@@ -174,28 +174,38 @@ class NativeModel(object):
...
-                try:
-                    self._context = FilePathContext(
-                        os.path.join(path, filename))
-                    with open(self._context.filepath, 'r') as f:
-                        self._model = yaml_load(f)
-                        break
-                except:
-                    logger.debug('Failed to load model file', exc_info=True)
+        if type(model_from) is str:
+            if '\n' in model_from:
```
jonls: This is unfortunately not a reliable way of checking if a string is a path or a YAML document. It is possible to create a valid YAML document that has no newline characters.

psamm/datasource/native.py
```diff
@@ -174,28 +174,38 @@ class NativeModel(object):
...
-            for filename in DEFAULT_MODEL:
-                try:
-                    self._context = FilePathContext(
-                        os.path.join(path, filename))
-                    with open(self._context.filepath, 'r') as f:
-                        self._model = yaml_load(f)
-                        break
-                except:
-                    logger.debug('Failed to load model file', exc_info=True)
+        if type(model_from) is str:
```
jonls: Checking a type should usually be done with `if isinstance(<var>, <types>)...` instead of `if type(<var>) is <type>...` (https://stackoverflow.com/questions/21894575/isinstancefoo-bar-vs-typefoo-is-bar). This also applies to the other type checks below. Also, checking string types should use `string_types` from the `six` module instead of `str` because of the changes between Python 2 and 3.

psamm/datasource/native.py
```diff
@@ -174,28 +174,38 @@ class NativeModel(object):
...
+                        logger.debug('Failed to load model file',
+                                     exc_info=True)
+                else:
+                    # No model could be loaded
+                    raise ParseError('No model file could be found({})'.format(
+                        ', '.join(DEFAULT_MODEL)))
+        elif type(model_from) is dict:
+            self._context = None
+            self._model = model_from
+        elif type(model_from) is file:
```
jonls: Usually, it is not a good idea to determine if a variable is a file based on the type. There exists alternative file-like objects that behave exactly like a file but is not `file` type and not even a type that is a subclass to `file` (so `isinstance` does not help either). Instead this should be handled by duck-typing or checking explicitly that an attribute is available. For example, to check whether a variable holds a file-like object that we can read from, we can check that it has a `read` method: `if hasattr(model_from, 'read') and callable(model_from.read):...` or we can try to read from it and catch the exception if it fails (if there is a specific exception for this).

psamm/datasource/native.py
```diff
@@ -413,18 +423,6 @@ def parse_compound_yaml_file(path, f):
...
-    """
-    if format is None:
-        if (re.match(r'.+\.(yml|yaml)$', path)):
-            return 'yaml'
-        elif (re.match(r'.+\.tsv$', path)):
-            return 'tsv'
-    else:
-        return format.lower()
-
-
```
jonls: This change and the following ones revert the `resolve_format` changes from another pull request. This must be a mistake.

jonls:
First commit is good but needs a shorter commit message 👍 Try to think of the first commit as independent from the other changes. The commit message doesn't have to talk about `include`s in model, the change is just to allow a null context that doesn't do anything.

psamm/tests/test_yaml.py
```diff
@@ -114,6 +114,97 @@ def write_model_file(self, filename, contents):
...
+            limits:
+              - reaction: rxn_2_\u03c0
+                upper: 100
+            '''
+        m = native.NativeModel(longString)
+        with self.assertRaises(context.ContextError):
+            list(m.parse_medium())
+
+    def test_dict_model(self):
+        longString = '''---
```
jonls: Is this test supposed to test loading from dictionary? It seems to load from a string like above ;)

psamm/tests/test_yaml.py
```diff
@@ -114,6 +114,97 @@ def write_model_file(self, filename, contents):
...
+            compounds:
+              - id: A_\u2206
+              - id: B
+              - id: C
+            media:
+              - include: medium.yaml
+            limits:
+              - reaction: rxn_2_\u03c0
+                upper: 100
+            '''
```
jonls: Maybe this model could be more minimal? For example only include `name`, `biomass` and `reactions` and then put the `include` in the `reactions` list. The test would be on `parse_reactions()` instead. This is not strictly necessary but I think it would make the test easier to understand.

psamm/datasource/native.py
```diff
@@ -174,28 +174,39 @@ class NativeModel(object):
...
-                    self._context = FilePathContext(
-                        os.path.join(path, filename))
-                    with open(self._context.filepath, 'r') as f:
-                        self._model = yaml_load(f)
-                        break
-                except:
-                    logger.debug('Failed to load model file', exc_info=True)
+    def __init__(self, model_from):
+        """Create a model from the specified model file, directory,
+            long string, open file, or dictionary"""
```
jonls: Docstrings should be formatter according to https://www.python.org/dev/peps/pep-0257/ (it's similar to git commit messages). First line should be a summary (no line breaks!), then a blank line, then a longer description if necessary.

psamm/tests/test_yaml.py
```diff
@@ -114,6 +114,97 @@ def write_model_file(self, filename, contents):
             f.write(contents)
         return path
 
+    def test_long_string_model(self):
+        longString = '''---
```
jonls: We don't run flake on the tests but the usual style should still be followed when possible. Specifically, variables should be `snake_case` not `camelCase`.

psamm/tests/test_yaml.py
```diff
@@ -114,6 +114,97 @@ def write_model_file(self, filename, contents):
...
+                  - id: A_\u2206
+                  - id: C
+            limits:
+              - reaction: rxn_2_\u03c0
+                upper: 100
+            '''
+        m = native.NativeModel(longString)
+        self.assertEqual(m.get_name(), 'Test model')
+        self.assertEqual(m.get_biomass_reaction(), 'rxn_1')
+        self.assertEqual(m.get_extracellular_compartment(), 'e')
```
jonls: We could do some more testing here, for example, check that the reactions are parsed correctly, or at least check that the number of reactions is correct.

psamm/datasource/native.py
```diff
@@ -174,28 +174,39 @@ class NativeModel(object):
...
+        if isinstance(model_from, string_types):
+            if 'name: ' in model_from:
```
jonls: This is clever but still not safe... The `name:` may not appear in the string even though it is a model. I don't know if there is really a good solution to allowing a string to work as either an inline YAML or a path to a YAML file. It may be necessary to split off the file loading to (perhaps) a factory method?

psamm/datasource/native.py
```diff
@@ -174,28 +174,44 @@ class NativeModel(object):
...
+        """
+        if isinstance(model_from, string_types):
+            self._model = yaml_load(model_from)
+            self._context = context
+        elif isinstance(model_from, dict):
+            self._context = context
+            self._model = model_from
+        elif hasattr(model_from, 'read') and callable(model_from.read):
+            self._context = context
+            self._model = yaml_load(model_from)
```
jonls: What happens if I try to construct a `NativeModel` from an integer: `NativeModel(42)`? Hint: it should probably raise a `ValueError` in that case.

psamm/datasource/native.py
```diff
@@ -174,28 +174,44 @@ class NativeModel(object):
...
                 except:
-                    logger.debug('Failed to load model file', exc_info=True)
+                    logger.debug('Failed to load model file',
+                                 exc_info=True)
             else:
                 # No model could be loaded
-                raise ParseError('No model file could be found ({})'.format(
+                raise ParseError('No model file could be found({})'.format(
                     ', '.join(DEFAULT_MODEL)))
+        return model
```
jonls: Same here: what happens if the provided `path` does not exist? It will probably return `None` but it should preferably raise an exception.

psamm/datasource/native.py
```diff
@@ -174,28 +174,44 @@ class NativeModel(object):
...
+        elif isinstance(model_from, dict):
+            self._context = context
+            self._model = model_from
+        elif hasattr(model_from, 'read') and callable(model_from.read):
+            self._context = context
+            self._model = yaml_load(model_from)
+
+    @classmethod
+    def load_model_from_path(cls, path):
         if os.path.isfile(path):
```
jonls: It may be better to simply try to open as file (in a `try:` block) then catch the `IOError` if it fails and open as a directory instead. This is almost always better than using `os.path.isfile()` because it avoids a race condition.

PR #113, opened 2016-08-11
sbml: writer now uses NativeModel instead of MetabolicModel

freestraws:
continuation of https://github.com/zhanglab/psamm/pull/97

### DIFF
--- a/psamm/datasource/sbml.py
+++ b/psamm/datasource/sbml.py
@@ -648,7 +648,7 @@ class SBMLWriter(object):
         self._namespace = SBML_NS_L3_V1_CORE
         self._sbml_tag = partial(_tag, namespace=self._namespace)
 
-    def write_model(self, file, model, reactions, compounds):
+    def write_model(self, file, model):
         """Write a given model to file"""
 
         ET.register_namespace('mathml', MATHML_NS)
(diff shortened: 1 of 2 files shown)


psamm/datasource/sbml.py
```diff
@@ -713,20 +732,40 @@ def write_model(self, file, model, reactions, compounds):
 
         # Create list of reactions
         reactions = ET.SubElement(model_tag, self._sbml_tag('listOfReactions'))
-        for reaction in model.reactions:
+        # Create mapping for reactions containing flux limit definitions
+        for flux_limit in model.parse_limits():
```
jonls: You can rewrite this loop to `for reaction_id, lower, upper in model.parse_limits():...` to give the more clear variable names instead of `flux_limits[0]`, `flux_limits[1]`, etc.

psamm/datasource/sbml.py
```diff
@@ -713,20 +732,40 @@ def write_model(self, file, model, reactions, compounds):
...
+            eq_id = flux_limit[0]
```
jonls: I would change this from `eq_id` to `reaction_id`. It's more consistent with the naming in other parts of PSAMM. We don't use the term _equation ID_ elsewhere but we use _reaction ID_.

psamm/datasource/sbml.py
```diff
@@ -713,20 +732,40 @@ def write_model(self, file, model, reactions, compounds):
...
+            # Deal with lower limit
+            if flux_limit[1] is None:
+                lower = (
+                    0
+                    if reaction_equations[eq_id].direction == Direction.Forward
+                    else -model.get_default_flux_limit())
```
jonls: Using the inline `if` expression for this seems a bit odd because it is split over four lines. A normal `if` statement would probably be more clear.

psamm/datasource/sbml.py
```diff
@@ -736,11 +775,11 @@ def write_model(self, file, model, reactions, compounds):
                     self._sbml_tag('stoichiometry'), text_type(abs(value)))
 
             notes_tag = ET.SubElement(reaction_tag, self._sbml_tag('notes'))
-            if reaction in reaction_genes:
+            if eq_id in reaction_genes:
```
jonls: Use `reaction_id` instead of `eq_id` for the same reason as stated above.

psamm/datasource/sbml.py
```diff
@@ -750,8 +789,18 @@ def write_model(self, file, model, reactions, compounds):
...
-            lower_str = text_type(model.limits[reaction].lower)
-            upper_str = text_type(model.limits[reaction].upper)
+            if eq_id in flux_limits:
+                lower_str = text_type(flux_limits[eq_id][0])
+                upper_str = text_type(flux_limits[eq_id][1])
+            else:
+                lower_str = text_type(
+                    0
+                    if equation.direction == Direction.Forward
+                    else -model.get_default_flux_limit())
```
jonls: This duplicates the code from above. Ideally we should only have to do this conversion of `None` values to the default flux limit in one place.

jonls:
After you address the comments above, do the test with the model export again to see that it is identical.

psamm/datasource/sbml.py
```diff
@@ -713,20 +732,24 @@ def write_model(self, file, model, reactions, compounds):
 
         # Create list of reactions
         reactions = ET.SubElement(model_tag, self._sbml_tag('listOfReactions'))
-        for reaction in model.reactions:
+        # Create mapping for reactions containing flux limit definitions
+        for rxn_id, lower_lim, upper_lim in model.parse_limits():
+            eq_id = rxn_id
+            flux_limits[eq_id] = lower_lim, upper_lim
```
jonls: Just use `rxn_id` directly?

jonls:
Good, the only thing remaining is to take care of only writing out reactions in the model subset obtained from `NativeModel` through `parse_model()` (also use `has_model_definition()` to check whether a model subset is defined).

jonls:
The commit message on the last commit is not accurate anymore. Maybe just squash the two commits? Let me know when when it's ready for merge.

PR #115, opened 2016-08-19
sbml: compounds and reactions get unique, safe ids from their yaml model id

freestraws:


psamm/datasource/sbml.py
```diff
@@ -648,6 +649,44 @@ def __init__(self):
         self._namespace = SBML_NS_L3_V1_CORE
         self._sbml_tag = partial(_tag, namespace=self._namespace)
 
+    def _make_safe_id(self, thisid):
+        """Makes sure the ID only has the allowed characters.
```
jonls: Usually we should prefer be explicit about whether the function returns a value and what it returns, e.g.: "Return a modified id that has been made safe for SBML". Also, call the parameter `id`.

psamm/datasource/sbml.py
```diff
@@ -648,6 +649,44 @@ def __init__(self):
...
+        thisid = re.sub(r'\(([a-z])\)$', '_\\1', thisid)
+        thisid = thisid.replace('(', '_LPAREN_')
+        thisid = thisid.replace(')', '_RPAREN_')
+        thisid = thisid.replace('[', '_LSQBKT_')
+        thisid = thisid.replace(']', '_RSQBKT_')
+        thisid = thisid.replace(',', '_COMMA_')
+        thisid = thisid.replace('.', '_PERIOD_')
+        thisid = thisid.replace("'", '_APOS_')
+        thisid = re.sub(r'[^a-zA-Z0-9_]', '', thisid)
+        return thisid
```
jonls: I would chain the calls to `replace` to avoid all those duplicate `thisid`s, maybe like this:

``` python
id = (id.replace('-', '_DASH_')
        .replace('/', '_FSLASH_')
        .replace('\\', '_BSLASH_')
        ...)
```

You can do the `re.sub()` first and last. Or you could even do something like

``` python
substitutions = {
    '-': '_DASH_',
    '/': '_FSLASH_',
    ...
}
for symbol, escape in iteritems(substitutions):
    id = id.replace(symbol, escape)
```

psamm/datasource/sbml.py
```diff
@@ -648,6 +649,44 @@ def __init__(self):
...
+    def _create_unique_id(self, in_dict, orig_id):
+        """Creates a unique ID for reaction or compound.
+
+        Looks at the ID the thing would have and checks if it is a duplicate
+        of another in the dictionary and makes a new ID if so.
+        """
+
+        if orig_id in in_dict:
+            suffix = 1
+            while(True):
```
jonls: `while True:`

psamm/datasource/sbml.py
```diff
@@ -747,8 +796,10 @@ def write_model(self, file, model):
         reactions = ET.SubElement(model_tag, self._sbml_tag('listOfReactions'))
         for eq_id, equation in sorted(iteritems(reaction_equations)):
             reaction_tag = ET.SubElement(reactions, self._sbml_tag('reaction'))
-            reaction_tag.set(self._sbml_tag('id'), next(reaction_id))
-            reaction_tag.set(self._sbml_tag('name'), eq_id)
+            # eq_id = checkid(eq_id)
```
jonls: This comment is not needed.

PR #119, opened 2016-08-23
Release 0.24

freestraws:


NEWS.md
```diff
@@ -1,3 +1,11 @@
+v0.24 (2016-08-22)
+------------------
+- The `fixed` key is now recognized in limits
```
jonls: I would expand on this a bit more, something like: "When specifying flux bounds in `media` and `limits`, the `fixed` key can now be used when lower and upper limits are the same."

NEWS.md
```diff
@@ -1,3 +1,11 @@
...
+- new column in output of `excelexport` and `tableexport` commands to
+  indicate if reaction of compounds are in the model
```
jonls: Capitalize sentence. "reaction _or_ compounds"?

NEWS.md
```diff
@@ -1,3 +1,11 @@
...
+- Can input long strings, dicts, or open files as model
```
jonls: Internal change. I would omit this.

NEWS.md
```diff
@@ -1,3 +1,11 @@
...
+- Do not return values for zero mass compounds
```
jonls: "Zero mass compounds are now omitted from the output of the `massconsistency` command."

jonls:
Capitalize all sentences and end with period. Also mention that the SBML format changed (safe IDs now generated from YAML ID; also you can mention that the encoding is COBRA compatible).

jonls:
Merged with small change to NEWS in a61279717d950701e81070ff71efd70dd79ae9ba

PR #121, opened 2016-08-26
native: nonextracellular compounds of none compartment get default compartment

freestraws:


jonls:
Thanks 👍, can you please squash these?

PR #122, opened 2016-08-31
update docs with new release's changes

freestraws:


jonls:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;V4xqrvNQ&#x2F;30-go-through-documentation-especially-file-format-and-tutorial-to-make-sure-that-it-is-up-to-date"> Go through documentation (especially file format) and tutorial to make sure that it is up-to-date</a>

docs/file_format.rst
```diff
@@ -38,6 +38,13 @@ The optional ``extracellular`` key specifies the default string for
 the extracellular compartment on compounds. If this option is not
 specified it will be assumed that the extracellular compartment is called ``e``.
 
+-Default Compartment
+--------------------------
+
+The optional ``default`` key specifies the default string for
+the default compartment on compounds. If this option is not
```
jonls: Maybe we can change this a bit to clarify what it does: "The optional ``default`` key specifies the default compartment that is used if a compound in a reaction does not explicitly specify a compartment." Also, shouldn't this be `default_compartment`? I think `default` is too ambiguous as a key.

jonls: Also, I think we should provide an example. Something like: "For example, the reaction `|x[e]| + |atp| => |x| + |adp| + |pi|` does not specify a compartment on four of the compounds so these would automatically be presumed to appear in the default compartment (or `c` if no default compartment is specified)."

jonls:
The first commit includes a section on default compartment but the commit message says that it only adds info on the `fixed` key.

docs/file_format.rst
```diff
@@ -38,6 +38,16 @@ The optional ``extracellular`` key specifies the default string for
 the extracellular compartment on compounds. If this option is not
 specified it will be assumed that the extracellular compartment is called ``e``.
 
+-Default Compartment
+--------------------------
+
+The optional ``default compartment`` key specifies the default compartment
+that is used if a compound in a reaction does not explicitly specify a
+compartment.  For example, the reaction |x[e]| + |atp| => |x| + |adp| + |pi|
```
jonls: The reaction should be ``` `` ```-quoted so it is displayed as fixed-width font. I don't think the `default compartment` is correct, there should be an underscore in there. Did you check with the code that it actually parses `default_compartment` and not `default`?

docs/commands.rst
```diff
@@ -389,6 +389,23 @@ Exports the model to the SBML file format.
...
+    $ psamm-model excelexport model.xls
+
+Table Export (``tableexport``)
+-----------------------------
+
+Exports the model to the tsv file format.
+
+.. code-block:: shell
+
+    $ psamm-model tableexport reactions > model.tsv
```
jonls: Needs a blank line after this paragraph.

docs/tutorial.rst
```diff
@@ -336,15 +337,12 @@ give the basic statistics of the model and should look like this:
...
-    Model: Ecoli_core_model
-    - Biomass reaction: Biomass_Ecoli_core_w_GAM
-    - Compounds: 72
+    Model: Escerichia coli Textbook (core) model (Excel format), Orth et al., 2010
+    - Biomass reaction: None
+    - Compounds: 54
     - Reactions: 95
     - Genes: 137
-    INFO: Using default flux limit of 1000.0
+    INFO: e is extracellular compartment
```
jonls: You are running this with excel importer even though the command above is for sbml. You need to run this with the command above to see if the output should be changed.

docs/file_format.rst
```diff
@@ -38,6 +38,16 @@ The optional ``extracellular`` key specifies the default string for
 the extracellular compartment on compounds. If this option is not
 specified it will be assumed that the extracellular compartment is called ``e``.
 
+Default Compartment
+--------------------------
+
+The optional ``default compartment`` key specifies the default compartment
```
jonls: There should be an underscore in `default_compartment`.

PR #123, opened 2016-08-31
Sbmlwriter complies with fbcv2 guidelines

freestraws:


freestraws:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;E2Axdt4u&#x2F;31-support-fbcv2-in-sbmlexport"> Support FBCv2 in sbmlexport</a>

jonls:
Looks good. Can you do separate commits for independent changes? It would be good to have flux bound, objectives and genes in separate commits.

psamm/datasource/sbml.py
```diff
@@ -711,11 +712,14 @@ def write_model(self, file, model):
...
         reaction_names = {}
+        biomass_id = ''
         for r in model.parse_reactions():
             if (model_reactions is not None and
                     r.id not in model_reactions):
                 continue
-
+            if r.id == model.get_biomass_reaction():
+                biomass_id = self._create_unique_id(
+                    reaction_equations, self._make_safe_id(r.id))
```
jonls: You are generating a new ID so I don't think this will work correctly. The objective has to reference the same ID as you are using for the reaction.

psamm/datasource/sbml.py
```diff
@@ -711,11 +712,14 @@ def write_model(self, file, model):
         reaction_genes = {}
         reaction_equations = {}
         reaction_names = {}
+        biomass_id = ''
```
jonls: What if biomass is not set in the YAML model? Maybe this should be `None` by default?

psamm/datasource/sbml.py
```diff
@@ -807,6 +826,50 @@ def write_model(self, file, model):
...
+                    lower_str = text_type(-model.get_default_flux_limit())
+            else:
+                lower_str = text_type(flux_limits[eq_id][0])
+            if eq_id not in flux_limits or flux_limits[eq_id][1] is None:
+                if equation.direction == Direction.Reverse:
+                    upper_str = text_type(0)
+                else:
+                    upper_str = text_type(model.get_default_flux_limit())
+            else:
+                upper_str = text_type(flux_limits[eq_id][1])
```
jonls: We have code exactly like this in a few places now. Can you add a task on the trello board to refactor this?

jonls:
We also have to set `strict` on the model, right?

psamm/datasource/sbml.py
```diff
@@ -30,7 +30,7 @@
 from six import itervalues, iteritems, text_type
 
 from ..reaction import Reaction, Compound, Direction
-
+from ..expression.boolean import Expression, And, Or, Variable
```
jonls: This belongs in genes commit.

psamm/datasource/sbml.py
```diff
@@ -868,5 +894,16 @@ def write_model(self, file, model):
                 self._sbml_tag('constant'): 'true'
             })
 
+        # list all gene products
+        listallgenes = ET.SubElement(model_tag, _tag(
+            'listOfGeneProducts', FBC_V2))
```
jonls: Generally variable names should be `snake_case` instead of running all words together.

psamm/datasource/sbml.py
```diff
@@ -798,6 +803,21 @@ def write_model(self, file, model):
         for rxn_id, lower_lim, upper_lim in model.parse_limits():
             flux_limits[rxn_id] = lower_lim, upper_lim
 
+        # Objective
+        objectivelist = ET.SubElement(model_tag, _tag(
```
jonls: `snake_case`

psamm/datasource/sbml.py
```diff
@@ -798,6 +803,21 @@ def write_model(self, file, model):
...
+        objective_tag.set(_tag('id', FBC_V2), 'O_1')
+        objective_tag.set(_tag('type', FBC_V2), 'maximize')
+        fluxobjectivelist = ET.SubElement(objective_tag, _tag(
+            'listOfFluxObjectives', FBC_V2))
+        fluxobjective_tag = ET.SubElement(fluxobjectivelist, _tag(
+            'fluxObjective', FBC_V2))
+        fluxobjective_tag.set(_tag('reaction', FBC_V2), biomass_id)
+        fluxobjective_tag.set(_tag('coefficent', FBC_V2), '1')
+
+        gene_list = set()
```
jonls: This belongs in the gene commit, I think? Also, probably a bit confusing calling this `gene_list` when it is a set.

psamm/datasource/sbml.py
```diff
@@ -712,13 +712,18 @@ def write_model(self, file, model):
...
-
-            reaction_id = self._create_unique_id(
-                reaction_equations, self._make_safe_id(r.id))
+            if r.id == model.get_biomass_reaction():
+                biomass_id = self._create_unique_id(
+                    reaction_equations, self._make_safe_id(r.id))
+                reaction_id = biomass_id
+            else:
+                reaction_id = self._create_unique_id(
+                    reaction_equations, self._make_safe_id(r.id))
```
jonls: I think this is correct now, but why not just:
``` python
reaction_id = self._create_unique(...)
if r.id == model.get_biomass_reaction():
    biomass_id = reaction_id
```

psamm/datasource/sbml.py
```diff
@@ -798,6 +803,21 @@ def write_model(self, file, model):
...
             flux_limits[rxn_id] = lower_lim, upper_lim
 
+        # Objective
+        objectivelist = ET.SubElement(model_tag, _tag(
+            'listOfObjectives', FBC_V2))
+        objectivelist.set(_tag('activeObjective', FBC_V2), 'O_1')
+        objective_tag = ET.SubElement(objectivelist, _tag('objective', FBC_V2))
+        objective_tag.set(_tag('id', FBC_V2), 'O_1')
+        objective_tag.set(_tag('type', FBC_V2), 'maximize')
+        fluxobjectivelist = ET.SubElement(objective_tag, _tag(
```
jonls: `snake_case`

jonls:
The `write_model()` method is becoming huge... maybe split it into smaller methods? Or at least add the new additions as smaller methods? E.g. `_add_fbc_flux_bounds()`, `_add_fbc_objective()`, `_add_fbc_gene_product_list()`, `_add_fbc_gene_associations()`...

psamm/datasource/sbml.py
```diff
@@ -692,6 +692,23 @@ def _create_unique_id(self, in_dict, orig_id):
                 suffix += 1
         return orig_id
 
+    def _add_flux_bounds(self, r_id, model, flux_limits, equation):
```
jonls: Doesn't `add` flux bounds but returns something... maybe call it `_get_flux_bounds` or change it to actually set the `lowerFluxBound`, `upperFluxBound` properties.

psamm/datasource/sbml.py
```diff
@@ -692,6 +692,23 @@ def _create_unique_id(self, in_dict, orig_id):
...
+            lower_str = text_type(flux_limits[r_id][0])
+        if r_id not in flux_limits or flux_limits[r_id][1] is None:
+            if equation.direction == Direction.Reverse:
+                upper_str = text_type(0)
+            else:
+                upper_str = text_type(model.get_default_flux_limit())
+        else:
+            upper_str = text_type(flux_limits[r_id][1])
+        return lower_str, upper_str
+
```
jonls: Now that I look at this again I start to think that this is incorrect according to the FBCv2 spec. The `lowerFluxBound` and `upperFluxBound` should not contain the actual values but should instead contain references to the parameters defined below. Can you comment on that?

psamm/datasource/sbml.py
```diff
@@ -709,11 +709,51 @@ def _add_flux_bounds(self, r_id, model, flux_limits, equation):
...
+                current, parent = gene_stack.pop()
+                if isinstance(current, Or):
+                    gene_tag = ET.SubElement(parent, _tag('or', FBC_V2))
+                elif isinstance(current, And):
+                    gene_tag = ET.SubElement(parent, _tag('and', FBC_V2))
+                elif isinstance(current, Variable):
+                    gene_tag = ET.SubElement(parent, _tag(
+                        'geneProductRef', FBC_V2))
+                    gene_tag.set(_tag('geneProduct', FBC_V2),
+                                 current.symbol)
```
jonls: I don't think you can use `current.symbol` directly here. `current.symbol` is the label of the gene but this needs to be the SBML id for the gene. You seem to be generating the unique IDs in `_add_gene_list` so you need to supply that ID mapping to this method or work around it in some other way.

psamm/datasource/sbml.py
```diff
@@ -709,11 +709,51 @@ def _add_flux_bounds(self, r_id, model, flux_limits, equation):
...
+            gene_ids[gene] = id
+            gene_tag.set(_tag('id', FBC_V2), 'g_' + id)
+            gene_tag.set(_tag('label', FBC_V2), gene)
+
     def write_model(self, file, model):
         """Write a given model to file"""
 
         ET.register_namespace('mathml', MATHML_NS)
         ET.register_namespace('xhtml', XHTML_NS)
+        ET.register_namespace('fbc', FBC_V2)
```
jonls: This has to go in the first commit. Either a separate commit that only adds this namespace or the first of your commits that uses the namespace.

psamm/datasource/sbml.py
```diff
@@ -709,6 +709,21 @@ def _add_flux_bounds(self, r_id, model, flux_limits, equation):
             upper_str = text_type(flux_limits[r_id][1])
         return lower_str, upper_str
 
+    def _add_fbc_objective(self, m_tag, fbc, obj_id):
```
jonls: Add docstrings to these methods. You can also try to improve the parameter names a bit. Maybe `model_tag` is better than `m_tag`? Maybe you don't have to pass `fbc`?

psamm/datasource/sbml.py
```diff
@@ -807,6 +826,12 @@ def write_model(self, file, model):
             reaction_tag.set(self._sbml_tag('reversible'), text_type(
                 equation.direction == Direction.Both).lower())
 
+            lower_str, upper_str = self._get_flux_bounds(
+                eq_id, model, flux_limits, equation)
+
+            reaction_tag.set(self._sbml_tag('lowerFluxBound'), lower_str)
+            reaction_tag.set(self._sbml_tag('upperFluxBound'), upper_str)
```
jonls: Now that I look at this again I start to think that this is incorrect according to the FBCv2 spec. The `lowerFluxBound` and `upperFluxBound` should not contain the actual values but should instead contain references to the parameters defined below. Can you comment on that?

jonls: These attributes are also in the SBML namespace but I think they belong in the FBC namespace?

psamm/datasource/sbml.py
```diff
@@ -711,11 +711,53 @@ def _get_flux_bounds(self, r_id, model, flux_limits, equation):
...
+                current, parent = gene_stack.pop()
+                if isinstance(current, Or):
+                    gene_tag = ET.SubElement(parent, _tag('or', FBC_V2))
+                elif isinstance(current, And):
+                    gene_tag = ET.SubElement(parent, _tag('and', FBC_V2))
+                elif isinstance(current, Variable):
+                    gene_tag = ET.SubElement(parent, _tag(
+                        'geneProductRef', FBC_V2))
+                    if current.symbol not in gene_ids:
+                        # gene_set.add(current.symbol)
```
jonls: This comment should be removed if the code is not needed.

psamm/datasource/sbml.py
```diff
@@ -711,11 +711,53 @@ def _get_flux_bounds(self, r_id, model, flux_limits, equation):
...
+                        gene_ids[id] = current.symbol
+                    gene_tag.set(_tag('geneProduct', FBC_V2),
+                                 gene_ids[id])
+                if isinstance(current, (Or, And)):
+                    for item in current:
+                        gene_stack.append((item, gene_tag))
+        return gene_ids
+
+    def _add_gene_list(self, parent_tag, gene_id_list):
+        """ list all gene products"""
```
jonls: This docstring should be formatted properly.

psamm/datasource/sbml.py
```diff
@@ -711,11 +711,53 @@ def _get_flux_bounds(self, r_id, model, flux_limits, equation):
...
+            gene_tag = ET.SubElement(list_all_genes, _tag('gene', FBC_V2))
+            gene_tag.set(_tag('id', FBC_V2), id)
+            gene_tag.set(_tag('label', FBC_V2), gene_id_list[id])
+
     def write_model(self, file, model):
         """Write a given model to file"""
 
         ET.register_namespace('mathml', MATHML_NS)
         ET.register_namespace('xhtml', XHTML_NS)
+        ET.register_namespace('fbc', FBC_V2)
```
jonls: This has to go in the first commit. Either a separate commit that only adds this namespace or the first of your commits that uses the namespace.

freestraws: it is in the first commit that uses that namespace.

psamm/datasource/sbml.py
freestraws: which parameters do you mean by defined below?

psamm/datasource/sbml.py
freestraws: (it was in the first that used it until I changed fluxbounds to use it just now)

psamm/datasource/sbml.py
jonls: You define a parameter here called `LOWER_BOUND`: https://github.com/zhanglab/psamm/pull/123/commits/a167a0455e8f413c8abdcbe761b56cd4af4eb37c#diff-c8d49b0ea75d56cd1cb1d9648030872dL853 What I mean is that, the `lowerFluxBound` should not contain the numerical value of the lower bound directly but should instead refer to the parameter (i.e. `lowerFluxBound="LOWER_BOUND"`). You have to check this with the specification, though, to make sure. Also I'm not sure if it is valid to use the id `LOWER_BOUND` for every reaction. You may need to change the parameter ID to be unique (again, I'm not entirely sure about this so check with the specification).

psamm/datasource/sbml.py
jonls: 👍

psamm/datasource/sbml.py
```diff
@@ -701,14 +702,20 @@ def _get_flux_bounds(self, r_id, model, flux_limits, equation):
             else:
                 lower_str = text_type(-model.get_default_flux_limit())
         else:
-            lower_str = text_type(flux_limits[r_id][0])
+            lower = flux_limits[r_id][0]
+            if lower % 1 == 0 and lower != -model.get_default_flux_limit():
```
jonls: I don't get the second condition (`lower != -model.get_default_flux_limit()`)... why only convert to `int` if it is not at the limit?

psamm/datasource/sbml.py
```diff
@@ -872,12 +894,20 @@ def write_model(self, file, model):
             species_tag.set(
                 self._sbml_tag('compartment'),
                 model_compartments[species.compartment])
+            species_tag.set(self._sbml_tag('constant'), 'true')
+            species_tag.set(self._sbml_tag('boundaryCondition'), 'true')
```
jonls: Should always be `false`.

psamm/datasource/sbml.py
```diff
@@ -872,12 +894,20 @@ def write_model(self, file, model):
...
+            species_tag.set(self._sbml_tag('hasOnlySubstanceUnits'), 'true')
+
+        if self._cobra_flux_bounds is False:
```
jonls: `if not self._cobra_flux_bounds:`

psamm/datasource/sbml.py
```diff
@@ -888,20 +918,29 @@ def write_model(self, file, model):
                 reaction_tag.set(self._sbml_tag('name'), reaction_names[eq_id])
             reaction_tag.set(self._sbml_tag('reversible'), text_type(
                 equation.direction == Direction.Both).lower())
+            reaction_tag.set(self._sbml_tag('fast'), 'false')
 
             lower_str, upper_str = self._get_flux_bounds(
                 eq_id, model, flux_limits, equation)
 
+            if self._cobra_flux_bounds is False:
```
jonls: `if not ...`

psamm/datasource/sbml.py
jonls: Actually, maybe we should always generate `listOfParameters` since this can now coexist with the COBRA-style local parameters (right?).

psamm/datasource/sbml.py
freestraws: The default flux limit here is usually 1000.0 as far as I could see, so without this there would be duplicates.

psamm/datasource/sbml.py
freestraws: It said that if constant was true, boundary condition also had to be true (in that validator).

psamm/datasource/sbml.py
freestraws: that wouldn't be confusing for the user? ok, i guess

jonls: As long as the old and new parameters are consistent I don't think it will be an issue. SBML files are not usually edited directly by users. Also, the default should be to generate only the FBC parameter list and not the local parameters.

freestraws: Yep, that's the default right now

psamm/datasource/sbml.py
```diff
@@ -711,6 +711,46 @@ def _get_flux_bounds(self, r_id, model, flux_limits, equation):
...
+                        gene_ids[id] = current.symbol
+                    gene_tag.set(_tag('geneProduct', FBC_V2),
+                                 gene_ids[id])
+                if isinstance(current, (Or, And)):
+                    for item in current:
+                        gene_stack.append((item, gene_tag))
+        return gene_ids
+
+    def _add_gene_list(self, parent_tag, gene_id_list):
+        """ List of all gene products as sbml readable elements."""
```
jonls: Remove initial whitespace in docstring. I would change the name of `gene_id_list` since it is not a list (maybe `gene_id_dict`). Docstring should start with a verb (maybe `Create list of all gene product...`).

psamm/datasource/sbml.py
```diff
@@ -711,6 +711,46 @@ def _get_flux_bounds(self, r_id, model, flux_limits, equation):
...
+                    if current.symbol not in gene_ids:
+                        id = 'g_' + self._create_unique_id(
+                            gene_ids, self._make_safe_id(current.symbol))
+                        gene_ids[id] = current.symbol
+                    gene_tag.set(_tag('geneProduct', FBC_V2),
+                                 gene_ids[id])
+                if isinstance(current, (Or, And)):
+                    for item in current:
+                        gene_stack.append((item, gene_tag))
+        return gene_ids
```
jonls: You don't have to return this since you are modifying the argument `gene_ids`.

psamm/datasource/sbml.py
```diff
@@ -711,6 +711,46 @@ def _get_flux_bounds(self, r_id, model, flux_limits, equation):
...
+
+    def _add_gene_list(self, parent_tag, gene_id_list):
+        """ List of all gene products as sbml readable elements."""
+        list_all_genes = ET.SubElement(parent_tag, _tag(
+            'listOfGeneProducts', FBC_V2))
+        for id in sorted(gene_id_list):
```
jonls: Here I would use `iteritems` to iterate over key, value pairs instead of just keys. Then you don't need another lookup when you set the `label`.

psamm/datasource/sbml.py
```diff
@@ -768,6 +774,17 @@ def _add_gene_list(self, parent_tag, gene_id_list):
...
+    def _transform_param(self, stri):
+        if '.' in stri:
+            stri = ''+stri.replace('.', '_')
+        elif '_' in stri:
+            stri = ''+stri.replace('_', '.')
+        if '-' in stri:
+            stri = ''+stri.replace('-', 'neg')
+        elif 'neg' in stri:
+            stri = ''+stri.replace('neg', '-')
+        return stri
```
jonls: This seems strange. Why do you do both the forward and reverse conversions. Also, why `''+...`?

jonls: This method is not used in this commit...?

psamm/datasource/sbml.py
```diff
@@ -742,8 +742,7 @@ def _add_gene_associations(self, r_id, r_genes, gene_ids, r_tag):
                         id = 'g_' + self._create_unique_id(
                             gene_ids, self._make_safe_id(current.symbol))
                         gene_ids[id] = current.symbol
-                    gene_tag.set(_tag('geneProduct', FBC_V2),
-                                 gene_ids[id])
+                        gene_tag.set(_tag('geneProduct', FBC_V2), id)
```
jonls: Squash this change into the gene product commit?

psamm/datasource/sbml.py
```diff
@@ -762,15 +761,16 @@ def _add_fbc_objective(self, model_tag, obj_id):
             'listOfFluxObjectives', FBC_V2))
         flux_objective_tag = ET.SubElement(flux_objective_list, _tag(
             'fluxObjective', FBC_V2))
-        flux_objective_tag.set(_tag('reaction', FBC_V2), obj_id)
-        flux_objective_tag.set(_tag('coefficent', FBC_V2), '1')
+        flux_objective_tag.set(_tag('reaction', FBC_V2), 'R_' + obj_id)
+        flux_objective_tag.set(_tag('coefficient', FBC_V2), '1')
```
jonls: Squash this change into the objective commit?

psamm/datasource/sbml.py
```diff
@@ -762,15 +761,16 @@ def _add_fbc_objective(self, model_tag, obj_id):
...
 
     def _add_gene_list(self, parent_tag, gene_id_list):
         """ List of all gene products as sbml readable elements."""
         list_all_genes = ET.SubElement(parent_tag, _tag(
             'listOfGeneProducts', FBC_V2))
         for id in sorted(gene_id_list):
-            gene_tag = ET.SubElement(list_all_genes, _tag('gene', FBC_V2))
+            gene_tag = ET.SubElement(
+                list_all_genes, _tag('geneProduct', FBC_V2))
```
jonls: Squash this change into the gene commit?

psamm/datasource/sbml.py
```diff
@@ -842,6 +844,7 @@ def write_model(self, file, model):
         root = ET.Element(self._sbml_tag('sbml'))
         root.set(self._sbml_tag('level'), '3')
         root.set(self._sbml_tag('version'), '1')
+        root.set(_tag('required', FBC_V2), 'false')
```
jonls: Squash into the first commit that adds FBC namespace?

psamm/datasource/sbml.py
```diff
@@ -831,6 +831,8 @@ def write_model(self, file, model):
             if reaction_id is None:
                 reaction_id = 'EX_{}_{}'.format(
                     compound.name, compound.compartment)
+            reaction_id = self._create_unique_id(
+                reaction_equations, self._make_safe_id(reaction_id))
```
jonls: Squash into appropriate commit?

psamm/datasource/sbml.py
```diff
@@ -962,5 +983,11 @@ def write_model(self, file, model):
 
         self._add_gene_list(model_tag, gene_ids)
 
+        for p in params:
+            param_tag = ET.SubElement(params_list, self._sbml_tag('parameter'))
+            param_tag.set(self._sbml_tag('id'), 'P_'+self._transform_param(p))
+            param_tag.set(self._sbml_tag('value'), p)
+            param_tag.set(self._sbml_tag('constant'), 'true')
+
```
jonls: Maybe put the code for generating parameter list in a separate commit? And then just keep the validator fixes to the existing code in this commit?

psamm/datasource/sbml.py
```diff
@@ -773,6 +773,13 @@ def _add_gene_list(self, parent_tag, gene_id_dict):
             gene_tag.set(_tag('id', FBC_V2), id)
             gene_tag.set(_tag('label', FBC_V2), label)
 
+    def _transform_param(self, stri):
+        if '.' in stri:
+            stri = stri.replace('.', '_')
+        if '-' in stri:
+            stri = stri.replace('-', 'neg')
+        return stri
```
jonls: There's no point in doing `if character in stri` when you call `replace` afterwards. `replace` is safe to call even if the character is not in `stri`. This means that the method can be written simply as
``` python
def _transform_param(self, stri):
    return stri.replace('.', '_').replace('-', 'neg')
```

Also, maybe change the name to reflect that this method is specifically for flux bound parameters? or at least for numerical values... (maybe `_make_safe_numerical_id` or something like that)

psamm/datasource/sbml.py
```diff
@@ -907,6 +917,13 @@ def write_model(self, file, model):
             lower_str, upper_str = self._get_flux_bounds(
                 eq_id, model, flux_limits, equation)
 
+            params[upper_str] = self._transform_param(upper_str)
+            params[lower_str] = self._transform_param(lower_str)
```
jonls: You could prepend the `'P_' + ` here and then you don't have to do that in the three places below.

psamm/datasource/sbml.py
jonls: :+1: This should be the first commit.

jonls:
The commit "sbml: global list of parameters" should be squashed into the first commit that sets flux bounds on reactions.

psamm/datasource/sbml.py
```diff
@@ -894,20 +900,21 @@ def write_model(self, file, model):
...
+            reaction_tag.set(self._sbml_tag('fast'), 'false')
 
             lower_str, upper_str = self._get_flux_bounds(
                 eq_id, model, flux_limits, equation)
 
             self._add_gene_associations(
                 eq_id, reaction_genes, gene_ids, reaction_tag)
 
-            reaction_tag.set(_tag('lowerFluxBound', FBC_V2), lower_str)
-            reaction_tag.set(_tag('upperFluxBound', FBC_V2), upper_str)
```
jonls: These deletions here seem strange.

psamm/datasource/sbml.py
```diff
@@ -864,6 +883,7 @@ def write_model(self, file, model):
             flux_limits[rxn_id] = lower_lim, upper_lim
 
         gene_ids = {}
+        self._add_fbc_objective(model_tag, biomass_id)
```
jonls: This should only be called if `biomass_id is not None`.

Issue #124, opened 2016-09-02
YAML reaction incorrectly parsing decimal stoichiometry as float


jonls:
The following model fails when running FBA. It appears that the `0.001` stoichiometric value for `a` in the `biomass` is parsed as a `float` while the `0.5` stoichiometric value for `a` in `rxn_1` is parsed as `Decimal`. When running FBA, these two values should be added to construct the LP equations but `float` and `Decimal` cannot be added. This would be solved if the stoichiometric values for the YAML reaction (`biomass` in this case) were correctly parsed as `Decimal`.

``` yaml
biomass: biomass
reactions:
  - id: rxn_1
    equation: (0.5) a => (0.5) c
  - id: biomass
    equation:
      left:
        - id: a
          value: 0.001
      right:
        - id: b
          value: 0.002
```

jonls:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;9TjOvZL5&#x2F;35-fix-parser-error-with-yaml-reactions"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> Fix parser error with YAML reactions</a>

PR #126, opened 2016-09-19
native: psamm-model correctly imports decimal values (not as float)

freestraws:
https://github.com/zhanglab/psamm/issues/124

### DIFF
--- a/psamm/datasource/native.py
+++ b/psamm/datasource/native.py
@@ -32,6 +32,7 @@ import csv
 
 import yaml
 from six import string_types, iteritems
+from decimal import Decimal
 
 from ..reaction import Reaction, Compound, Direction
 from ..metabolicmodel import MetabolicModel
@@ -55,6 +56,10 @@ class ParseError(Exception):
     """Exception used to signal errors while parsing"""
 
 
+def float_constructor(loader, node):
+    return Decimal(loader.construct_scalar(node))
+
+
 def whendefined(func, value):
     """Apply func to value if value is not None"""
     return func(value) if value is not None else None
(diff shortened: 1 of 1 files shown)


freestraws:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;9TjOvZL5&#x2F;35-fix-parser-error-with-yaml-reactions"> Fix parser error with YAML reactions</a>

psamm/datasource/native.py
```diff
@@ -74,7 +79,9 @@ def yaml_load(stream):
                            ' speed up loading the model files.')
 
     if _HAS_YAML_LIBRARY:
-        return yaml.load(stream, Loader=yaml.CSafeLoader)
+        loader = yaml.CSafeLoader
+        loader.add_constructor('tag:yaml.org,2002:float', float_constructor)
+        return yaml.load(stream, loader)
```
jonls: This only works when `_HAS_YAML_LIBRARY` is true, I think. We need to also add the constructor when the `CSafeLoader` is not available and we fall back to using `yaml.safe_load()` (below). I think this would be easier if we always explicitly use the loader instead of using the `load()` or `safe_load()` wrapper. This way you can also add the constructor to the loader instance instead of the class. I think my comment on Trello explains this. To summarize: 1) Construct the loader explicitly (`loader = yaml.CSafeLoader(stream)`), 2) Add the constructor on the loader _instance_ (`loader.add_constructor(...)`), 3) Parse the stream and return the result (`return loader.get_data()`). When `_HAS_YAML_LIBRARY` is false we have to use the Python-based loader instead of `yaml.CSafeLoader`. I think it is called `yaml.SafeLoader`.

PR #129, opened 2016-10-03
fastgapfill: exchange reactions not included in the core subset by default

freestraws:


freestraws:
[FastGapFill command should by default not include exchange reactions in the core subset.](https://trello.com/c/EbvonqfM/36-fastgapfill-command-should-by-default-not-include-exchange-reactions-in-the-core-subset)

psamm/commands/fastgapfill.py
```diff
@@ -96,7 +96,11 @@ def reaction_key(r):
             penalties=penalties)
 
         epsilon = self._args.epsilon
-        core = set(self._mm.reactions)
+        coreall = set(self._mm.reactions)
+        core = set()
+        for r in coreall:
```
jonls: I don't think it is necessary to save the `coreall` in a temporary set. How about just `for r in self._mm.reactions:`?

psamm/commands/fastgapfill.py
```diff
@@ -96,7 +96,11 @@ def reaction_key(r):
...
+            if not(self._mm.is_exchange(r)):
```
jonls: `if not self._mm.is_exchange(r):`

PR #130, opened 2016-10-03
fastgapfill: option to choose a subset of reactions to use

freestraws:


freestraws:
[FastGapFill command should have an option to specify the core reaction subset to use.](https://trello.com/c/zCnOmoWj/37-fastgapfill-command-should-have-an-option-to-specify-the-core-reaction-subset-to-use)

PR #132, opened 2016-10-14
sbml: export charge and formula

freestraws:


freestraws:
Not done yet, just a placeholder to remember what I was doing for next time.  Charge and formula still aren't detected on import though they appear in the sbml file (with the compartment added to the end but that's a different issue that I should also fix here...).

PR #148, opened 2016-10-28
gapfill: weights option

freestraws:


psamm/commands/gapfill.py
```diff
@@ -83,7 +108,13 @@ def run(self):
 
         if len(blocked) > 0:
             # Add exchange and transport reactions to database
-            model_complete = self._mm.copy()
+            model_complete, weights = create_extended_model(
+                self._model,
+                db_penalty=self._args.db_penalty,
+                ex_penalty=self._args.ex_penalty,
+                tp_penalty=self._args.tp_penalty,
+                penalties=penalties)
```
jonls: The `weights` dictionary should be supplied to the `gapfill` function.

psamm/commands/gapfill.py
```diff
@@ -83,7 +108,13 @@ def run(self):
...
+
             logger.info('Adding database, exchange and transport reactions')
             model_complete.add_all_database_reactions(model_compartments)
```
jonls: This line and the following two lines should not be called since the additional reactions are already added by `create_extended_model()`.

jonls:
Continued in #169.

PR #151, opened 2016-10-31
gapfill: report likely failure in GapFind/GapFill

freestraws:


jonls:
After looking more into the GapFind/GapFill issues it seems that there are better ways of making the command more robust and generating proper warnings. This was continued in #154.

PR #170, opened 2016-11-19
sbml, excelexport: added git version to info

freestraws:


psamm/datasource/sbml.py
```diff
@@ -856,6 +858,8 @@ def write_model(self, file, model):
         root.set(self._sbml_tag('level'), '3')
         root.set(self._sbml_tag('version'), '1')
         root.set(_tag('required', FBC_V2), 'false')
+        if git_version is not None:
+            root.set(self._sbml_tag('psamm_version'), git_version)
```
jonls: Unfortunately we cannot just invent the tag `psamm_version` in standards compliant SBML. The best option is probably to put it in a notes section. If I remember correctly, the whole model allows a notes section (not specific to a reaction or species) that could be used.

PR #171, opened 2016-11-19
sbmlexport: user can optionally specify output file

freestraws:


freestraws:
i used http://effbot.org/zone/element-lib.htm#prettyprint for the indentation/ pretty-formatted output.

jonls:
@freestraws Did you try `nargs='?'` as mentioned above?

psamm/commands/sbmlexport.py
```diff
@@ -24,6 +24,15 @@
 class SBMLExport(MetabolicMixin, Command):
     """Export model as SBML file."""
 
+    @classmethod
+    def init_parser(cls, parser):
+        parser.add_argument(
+            'file', type=str, help='File path for writing the Excel workbook',
+            default='-')
```
jonls: Should call the super method as the last thing in this method. See other commands for example.

psamm/datasource/sbml.py
```diff
@@ -1020,4 +1035,5 @@ def write_model(self, file, model):
         self._add_gene_list(model_tag, gene_ids)
 
         tree = ET.ElementTree(root)
+        self._indent(root)
```
jonls: I would like to only apply this if an option `--pretty` is provided to the `sbmlexport` command.

docs/commands.rst
```diff
@@ -388,9 +388,17 @@ SBML Export (``sbmlexport``)
 
 Exports the model to the SBML file format.
 
+The below will export the model to the terminal
+
+.. code-block:: shell
+
+    $ psamm-model sbmlexport -
+
+Using `--dest` option allows saving this output to a file:
```
jonls: This was changed to positional argument.

psamm/commands/sbmlexport.py
freestraws: Ooooh, I was looking at `excelexport` for reference and it doesn't use `nargs` or its `init_parser` for some reason, I thought that was the way to go.

psamm/commands/sbmlexport.py
```diff
@@ -24,6 +24,19 @@
 class SBMLExport(MetabolicMixin, Command):
     """Export model as SBML file."""
 
+    @classmethod
+    def init_parser(cls, parser):
+        parser.add_argument(
+            'file', type=str, help='File path for writing the Excel workbook',
```
jonls: 'Excel workbook' -> 'SBML model'

jonls:
Merged with minor changes in #195

PR #181, opened 2016-12-16
Move gapfind to new command gapcheck

jonls:


psamm/commands/gapcheck.py
```diff
@@ -0,0 +1,74 @@
...
+logger = logging.getLogger(__name__)
+
+
+class GapCheckCommand(MetabolicMixin, SolverCommandMixin, Command):
+    """Check for compound production gaps in model."""
+    @classmethod
+    def init_parser(cls, parser):
+        parser.add_argument(
+            '--threshold', type=float, default=1e-5,
+            help='Threshold for compound production')
```
keitht547: Do we want to start using this threshold option instead of what we were calling epsilon?
It seems to still be treated the same way as epsilon was.

PR #191, opened 2017-01-11
Added Minimization of Metabolic Adjustments (MOMA)

bishopbrian25:
Added the MOMA algorithms in a python script called moma.py.

There are 4 implementation of MOMA:

- MOMA (qlp2): Standard euclidean MOMA
- lin_MOMA (lp2): Standard manhattan MOMA
- MOMA2 (l1): Jon variation of euclidean MOMA
- lin_MOMA2 (l2): Jon variation of manhattan MOMA

### DIFF
--- a/psamm/moma.py
+++ b/psamm/moma.py
@@ -136,6 +136,25 @@ class MOMAProblem(object):
             raise MOMAError('Unable to solve L1 MOMA: {}'.format(
                 result.status))
 
+    def minimize_l2(self, objective):
+        wt_obj = self._solve_fba(objective)
+
+        obj_expr = 0
+        for reaction in self._adjustment_reactions():
+            v_wt = self._v_wt(reaction)
+            v = self._v(reaction)
+            obj_expr += (v_wt - v)**2
+
+        self._prob.set_objective(obj_expr)
+
+        v_wt_obj = self._v_wt(objective)
+        with self.constraints(v_wt_obj == wt_obj):
+            result = self._solve(lp.ObjectiveSense.Minimize)
+
+        if not result:
+            raise MOMAError('Unable to solve L2 MOMA: {}'.format(
+                result.status))
+
     def get_flux(self, reaction):
         return self._prob.result.get_value(self._v(reaction))
 
--- a/psamm/moma.py
+++ b/psamm/moma.py
@@ -35,7 +35,7 @@ class ConstraintGroup(object):
     """Constraints that will be imposed on the model when solving.
 
     Args:
-        moma: MOMAProblem for the gene deletion problem.
+        moma: MOMAProblem object for the proposed constraints.
         *args: The constraints that are imposed on the model.
     """
     def __init__(self, moma, *args):
@@ -51,7 +51,7 @@ class ConstraintGroup(object):
         self.delete()
 
     def add(self, *args):
-        """Add a constraint to the model."""
+        """Add a constraints to the model."""
         self._constrs.extend(self._moma._prob.add_linear_constraints(*args))
 
     def delete(self):
@@ -60,11 +60,24 @@ class ConstraintGroup(object):
 
 
 class MOMAProblem(object):
-    """Model as a flux optimization problem with steady state assumption.
+    """Model as a flux optimization problem with minimal flux redistribution.
 
     Create a representation of the model as an LP optimization problem with
-    steady state assumption, i.e. the concentrations of compounds are always
-    zero.
+    steady state assumption and a minimal redistribution of metabolic fluxes
+    with respect to the wild type configuration.
+
+    The problem can be solved using any of the four MOMA variants described in
+    [Segre02]_ and [Mo09]_. MOMA is formulated to avoid the FBA assumption that
+    that growth efficiency has evolved to an optimal point directly following
+    model perturbation. MOMA finds the optimal solution for a model with
+    minimal flux redistribution with respect to the wild type flux
+    configuration.
+
+    MOMA is implemented with two variations of a linear optimization problem
+    (:meth:`.lin_moma` and :meth:`.lin_moma2`) and two variations of a
+    quadratic optimization problem (:meth:`.moma` and :meth:`.moma2`).
+    Further information on these methods can be found within their respective
+    documentation.
 
     The problem can be modified and solved as many times as needed. The flux
     of a reaction can be obtained after solving using :meth:`.get_flux`.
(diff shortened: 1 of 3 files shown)


jonls:
@bishopbrian25 Can you rebase this branch on top of the branch from #172? I would like to see what changed in your version from the previous version of the `moma` module.

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+
+        # Return the dictionary with all of the fluxes associated with the
+        # reaction id
+        # {'reaction1': 1000, 'reaction2': 2000, ...}
+        return fba_fluxes
+
+    def _get_minimal_flux_fba(self, wt_biomass, objective):
+        """Finds the FBA solution that minimizes all the flux values, while
+        still maximizing the biomass. This still might not find a unique
+        solution."""
```
jonls: This docstring should follow the "one-sentence-summary" rule. Something like
```
"""Find the FBA solution that minimizes all the flux values.

Maximizes the objective flux then minimizes all other fluxes while keeping the
objective flux at the maximum."""
```

Try to use the terms biomass and objective consistently instead of mixing them. In this module I think it's most appropriate to use objective instead of biomass since the user may not necessarily use the biomass reaction flux as the objective.

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+        wt_obj = self._v_wt(objective)
+        with self.constraints(wt_biomass == wt_obj):
+            self._prob.set_objective(self._v_wt.sum(self._all_reactions()))
```
jonls: Unfortunately this will not work since we have reversible reactions with negative fluxes. We are interested in minimizing the sum of absolute fluxes not just the sum of fluxes. See the `minimize_l1()` method in the `fluxanalysis` module (you'll need additional variables to do this).

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+            result = self._solve(lp.ObjectiveSense.Minimize)
+        return result
+
+    # Solves for the FBA biomass. Uses the result returned by _solve_fba and
+    # finds the value of the objective (biomass) function.
+    def _get_fba_biomass(self, objective):
+        """Returns the flux for the max biomass found by FBA."""
+        # Run FBA and store the result
```
jonls: It's good that you have a lot of comments in this module but some of them may be superfluous if the code itself is quite obvious. In this case I think it is clear from the following line that you're solving fba and storing the result so the comment is not necessary. As a general rule, avoid comments that describe what you are doing if it's obvious. Keep comments that describe why you are doing something since those are much more useful.

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
     def _adjustment_reactions(self):
+        """Returns a generator of all the non exchange reactions
+        in the model."""
         for reaction_id in self._model.reactions:
             if not self._model.is_exchange(reaction_id):
                 yield reaction_id
 
+    # Returns a generator of all the reactions ids
+    def _all_reactions(self):
+        """Reaturns a generator of all the reactions in the model."""
```
jonls: typo: "Returns a generator...". Actually PEP 257 (https://www.python.org/dev/peps/pep-0257/) recommends a style where the initial verb is imperative i.e. ("Return a ..." instead of "Returns a ..."). I don't care too much about this recommendation but if you want to get into the habit of following this style, you can change these.

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+    # finds the value of the objective (biomass) function.
+    def _get_fba_biomass(self, objective):
+        """Returns the flux for the max biomass found by FBA."""
+        # Run FBA and store the result
+        flux_result = self._solve_fba(objective)
+        # Pull out the biomass value
+        return flux_result.get_value(self._v_wt(objective))
+
+    def minimize_lp2(self, objective, wt_fluxes):
+        """Implemetnation of the lin_MOMA algorithm.
```
jonls: typo: "Implementation"

jonls: Maybe just call "linear MOMA" instead of "lin_MOMA".

jonls: Actually, describe what the method does not what it is, i.e. something like: "Solve the linear MOMA ...".

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+
+        We try to maximize biomass, by keeping the fluxes relitively the same
```
jonls: typo: "...relatively..."

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+        as the wildtype. This helps avoid the assumption that an organism will
+        perform optimally directly after removing a gene.
+
+        This might not be the optimal minimization solution for the wild type
+        and the knockout vectors.
```
jonls: This is good. I think we can expand a little bit on the issues with this implementation. Maybe say that the objective flux in the MOMA result depends on the input wild type flux values so if those are generated as an arbitrary FBA solution the MOMA objective flux could change if a different flux solution was obtained from FBA.

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+
+        Minimizes sum of |wild type - knockout|
+
+        Args:
+            objective: Biomass reaction for the model.
+            wt_fluxes: All the wt_fluxes found by FBA in the form of a
+                dictionary. Use :meth: _get_fba_flux(objective) to return
+                a dictionary.
```
jonls: I think it's great to have a recommendation here specifying how this input value can be obtained. However, the `_get_fba_flux()` method is private (because of the initial `_`) so this contradicts the message. Either, we should make it public by dropping the initial underscore or add a different recommendation that is public (I think making the method public makes sense). Also, keep in mind that the paper on MOMA suggests that the fluxes could also come from experimental results, i.e. the dictionary is not required to come from FBA it is only a suggestion. We may want to move the recommendation to use the FBA fluxes to a paragraph above and then just specify that the argument should be a dictionary of flux values.

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+    # Solves for the FBA biomass. Uses the result returned by _solve_fba and
+    # finds the value of the objective (biomass) function.
+    def _get_fba_biomass(self, objective):
+        """Returns the flux for the max biomass found by FBA."""
+        # Run FBA and store the result
+        flux_result = self._solve_fba(objective)
+        # Pull out the biomass value
+        return flux_result.get_value(self._v_wt(objective))
+
+    def minimize_lp2(self, objective, wt_fluxes):
```
jonls: For the method name, I think we should use the same names as we decided on for the command. The name "lp2" is just a name I used in the document I provided which has no meaning outside of that document so we should not use it in the published code.

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+                # Add the constraint that finds the optimal solution, such
+                # that the difference between the wildtype flux is similar
+                # to the knockout flux.
+                constr.add(
+                    z(f_reaction) >= f_value - v(f_reaction),
+                    f_value - v(f_reaction) >= -z(f_reaction))
+
+        # If we minimize the sum of the z vector then we will minimize
+        # the |vs_wt - vs| from above
+        self._prob.set_objective(z.sum(self._adjustment_reactions()))
```
jonls: You can use the `reactions` variable here instead of calling `self._adjustment_reactions()` again.

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+        # Store the result of minimizing the sum of z
+        result = self._solve(lp.ObjectiveSense.Minimize)
+
+        # Need to make sure we catch any problems that occur in the solver
+        if not result:
+            raise MOMAError('Unable to solve LP2 MOMA: {}'.format(
+                result.status))
+
+        # Go back and manually delete all the constaints
+        constr.delete()
```
jonls: If an exception is raised (either a `MOMAError` or another exception) we return early from the method, and the constraints will not be removed on the next solve. This means that when an exception is raised the `MOMAProblem`is left in an inconsistent state that can no longer be reused! This may cause incorrect results if you reuse the problem instance anyway. The solution is to wrap every line after `const = self.constraints()` in a `try`-block and put the `constr.delete()` in the `finally`-block. This way, the constraints will always be removed, even if an exception is raised and we ensure that the problem instance is always left in the correct state for reuse.

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+
+    def minimize_l1(self, objective, wt_obj, wt_fluxes):
+        """Implemetnation of the lin_MOMA2 algorithm.
```
jonls: typo: "Implementation"...

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+
+        We try to maximize biomass, by keeping the fluxes relitively the same
+        as the wildtype. This helps avoid the assumption that an organism will
+        perform optimally directly after removing a gene.
```
jonls: Here you mention genes. I would leave this out and just talk about reactions since the gene removal is outside the scope of the MOMA algorithm.

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+
+        Creates the constraint that the we select the optimal flux vector that
+        is closest to the wildtype.
```
jonls: "...constraint that will select..."?

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+        # Set a constraint that says the biomass flux of the wild type is
+        # the same as the FBA analysis
         v_wt_obj = self._v_wt(objective)
         with self.constraints(v_wt_obj == wt_obj):
             result = self._solve(lp.ObjectiveSense.Minimize)
 
+        # Raise any errors
         if not result:
             raise MOMAError('Unable to solve L1 MOMA: {}'.format(
                 result.status))
```
jonls: Here, the constraints do not seem to be removed after solving. Should be handled with `try`-block as above.

psamm/moma.py
```diff
@@ -79,84 +100,332 @@ def __init__(self, model, solver):
...
+
+        # Set the objective
+        self._prob.set_objective(obj_expr)
+
+        # Minimize the squared distance between the wild type flux values and
+        # the knockout values
+        result = self._solve(lp.ObjectiveSense.Minimize)
+
+        # We need to go back and manually delete all the constaints
+        constr.delete()
```
jonls: Here the constraints are deleted before the `MOMAError` is potentially raised which is better and should in most cases produce the correct behavior when reusing the instance. Though, if an exception is thrown anywhere above we still get an inconsistent state. Use the `try`-block, 'finally'-block to handle this.

jonls:
@bishopbrian25 We should also add an index rst file in `docs/api/moma.rst` so that the `moma` module is included in the generated documentation. The file should contain:

```
``psamm.moma`` -- Minimization of metabolic adjustments
=======================================================

.. automodule:: psamm.moma
   :members:
```

jonls:
This was merged as part of #257

PR #192, opened 2017-01-11
Added MOMA implementation into genedelete command

bishopbrian25:
Added the MOMA implementation into the genedelete command. Also, added tests and documentation for the genedelete command.

Had to fix expression multiplication error and had to rewrite the lp test functions that was effected.

FBA is still the default use of genedelete and the arg "--method moma" can be used to use MOMA.

There are 5 --method options:
- fba
- moma
- moma2
- lin_moma
- lin_moma2


jonls:
Merged in #258

PR #196, opened 2017-01-16
Fixed --help newline styling

bishopbrian25:
When running a command with the --help option, the description of the command is shown. This description is based on the docstring of the Command subclass representing the running command. However, the formatting of the original docstring is not preserved. For example, the tableexport command contains a docstring with a list formatted with bullet points indentation and whitespace. In the output all newlines are converted to single spaces. The output should use the original formatting insert.

Fixed the problem by using RawDescriptionHelpFormatter as formatter_class. Description of the argument is below.

> Passing RawDescriptionHelpFormatter as formatter_class= indicates that description and epilog are already correctly formatted and should not be line-wrapped:

Documentation: [docs.python.org](https://docs.python.org/3/library/argparse.html#formatter-class)

### DIFF
--- a/psamm/command.py
+++ b/psamm/command.py
@@ -516,6 +516,7 @@ def main(command_class=None, args=None):
             title, _, _ = command_class.__doc__.partition('\n\n')
             subparser = subparsers.add_parser(
                 name, help=title.rstrip('.'),
+                formatter_class=argparse.RawDescriptionHelpFormatter,
                 description=command_class.__doc__)
             subparser.set_defaults(command=command_class)
             command_class.init_parser(subparser)
--- a/psamm/command.py
+++ b/psamm/command.py
@@ -459,6 +459,34 @@ class SequentialExecutor(Executor):
         pass
 
 
+def _trim(docstring):
+    """Return a trimmed docstring.
+
+    Code taken from 'PEP 257 -- Docstring Conventions' article.
+    """
+    if not docstring:
+        return ''
+    # Convert tabs to spaces (following the normal Python rules)
+    # and split into a list of lines:
+    lines = docstring.expandtabs().splitlines()
+    # Determine minimum indentation (first line doesn't count):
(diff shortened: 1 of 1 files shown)


jonls:
This is certainly a lot better than the output before. For example the `tableexport` command help now looks like this:

```
usage: psamm-model tableexport [-h] export_type

Export parts of the model as tab separated tables.

    This command can export various parts of the model as a TSV table.
    The output is written to standard output. The type argument is used
    to determine which part of the model to export:

    - reactions: Export reactions and reaction metadata
    - compounds: Export compounds and compound metadata
    - medium: Export the list of exchange compounds/reactions
    - limits: Export list of internal flux limits
    - metadata: Export general model metadata
    

positional arguments:
  export_type  Type of model data to export

optional arguments:
  -h, --help   show this help message and exit
```

instead of this

```
usage: psamm-model tableexport [-h] export_type

Export parts of the model as tab separated tables. This command can export
various parts of the model as a TSV table. The output is written to standard
output. The type argument is used to determine which part of the model to
export: - reactions: Export reactions and reaction metadata - compounds:
Export compounds and compound metadata - medium: Export the list of exchange
compounds/reactions - limits: Export list of internal flux limits - metadata:
Export general model metadata

positional arguments:
  export_type  Type of model data to export

optional arguments:
  -h, --help   show this help message and exit
```

Would it also be possible to remove the indentation of the paragraphs following the first one? In the example above the second paragraph and the list are both indented four spaces. Maybe replace any newline character followed by spaces with just a single newline? Also, we could perhaps add a newline at the end of the text if none is there? Currently some of the help texts have only one empty line between the text and "positional arguments" while others have two.

jonls:
Actually the recommended way of trimming (removing indentation of) a docstring is described in https://www.python.org/dev/peps/pep-0257/#handling-docstring-indentation. They even provide the code so it should be easy to implement.

PR #212, opened 2017-03-01
Rename media to exchange

jonls:
Rename the "media" key in the model file to "exchange" reflecting the fact that it specifies both compounds with allowed uptake (the medium) and compounds allowed to exit the model system. Also introduces related changes in `tableexport` and `excelexport` commands.

The old "media" key is still parsed in the model files but is deprecated and a warning is produced if it is present. The "medium" command in `tableexport` still works but the same command can also be run be specifying "exchange".

Documentation on the file format and tutorial is updated with this change.

- [x] Update `psamm-import` to use the new key (https://github.com/zhanglab/psamm-import/pull/35) (and import medium and exchange compounds to separate files?)
- [x] Update tutorial repository (https://github.com/zhanglab/psamm-tutorial/pull/1).

### DIFF
--- a/docs/file_format.rst
+++ b/docs/file_format.rst
@@ -31,8 +31,8 @@ used as a template:
       - include: reactions/reactions.tsv
       - include: reactions/biomass.yaml
 
-    media:
-      - include: medium.yaml
+    exchange:
+      - include: exchange.yaml
     limits:
       - include: limits.yaml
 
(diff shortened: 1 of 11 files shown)


jonls:
@keitht547 Can you please check whether the updates to the documentation look ok?

docs/file_format.rst
```diff
@@ -188,13 +188,20 @@ group of genes or when multiple genes can independently enable a reaction:
...
+------------------
 
-The optional ``media`` key provides a way of defining the medium (boundary
-conditions) for the model. The medium is defined by a set of compounds that are
-able enter or leave the model system. The following fragment is an example of
-the ``medium.yaml`` file:
+The ``exchange`` key provides a way of defining the compounds that can
+enter and exit the model system (the boundary conditions). This includes
+compounds that can enter the system (*the medium*) and compounds that are
+allowed to exit the system. In most cases, all compounds that occur in the
```
keitht547: maybe add in: "allowed to exit the system, like metabolic byproducts."

docs/tutorial.rst
```diff
@@ -709,13 +709,17 @@ non-default limit. This reaction is an ATP maintenance reaction and the
...
+Exchange compounds
+~~~~~~~~~~~~~~~~~~
 
-The medium file is where you can designate the boundary conditions for the
-model. The compartment of the medium compounds can be designated using the
+The exchange file is where you can designate the boundary conditions for the
+model. This includes both compounds that are allowed to enter the model system
+(*the medium*) as well as compounds that are allowed to exit the model.
+Settings a negative lower limit on an exchange compounds allows an uptake of
+that compound thereby making the compound part of the medium. The
```
keitht547: This might be confusing to say making it part of the medium. This sentence "Settings a negative lower limit on an exchange compounds allows an uptake of that compound thereby making the compound part of the medium." could probably be removed to avoid the confusion of saying some subset of the exchange will be called the medium.

jonls: I would like to explain that negative values correspond to uptake since I don't think we explain this anywhere else. There is also a sentence later in this document: "For example if glucose was added into the medium along with mannitol then the results might appear as follows" where "added into the medium" specifically means that the lower bound was set to a negative value. For that to make sense the medium has to be the set of exchange compounds with negative lower bound. Maybe we can change that sentence to spell out that the lower bound is set negative in the exchange file instead of saying that it was "added into the medium"? What do you think?

PR #262, opened 2017-06-26
Complete path

keitht547:
This implements a gapfilling extension that can be used to produce gapfilling results in the context of a flux consistent pathway for the production of a specified compound. This implementation of gapfilling varies from the 'psamm-model gapfill' command in its separation of artificial exchange reactions into distinct sink and source reactions with different penalties and in its ability to produce an FBA derived synthesis pathway for a compound.

### DIFF
--- a/psamm/commands/completepath.py
+++ b/psamm/commands/completepath.py
@@ -178,69 +178,69 @@ class CompletePathCommand(MetabolicMixin, SolverCommandMixin, Command):
                ' the model.'.format(epsilon))
         self.fail(msg, exc)
 
-    def pathway_extraction(metabolic_model, model_rxns, model_cpd_list, compound, solver):
-        '''Extract the pathway of reactions needed for the production of the compound after gapfilling.'''
 
-        for compound_id in compound:
-            model = metabolic_model.copy()
+def pathway_extraction(metabolic_model, model_rxns, model_cpd_list, compound, solver):
+    '''Extract the pathway of reactions needed for the production of the compound after gapfilling.'''
 
-            for rxn in metabolic_model.reactions:
-                if rxn not in model_rxns:
-                    model.remove_reaction(rxn)
-            reaction_id = 'tmp_objective'
-            reaction_ex = Reaction(Direction.Forward, {compound_id: -1})
-            model.database.set_reaction(reaction_id, reaction_ex)
-            model.add_reaction(reaction_id)
+    for compound_id in compound:
+        model = metabolic_model.copy()
 
-            prob = solver.create_problem()
+        for rxn in metabolic_model.reactions:
+            if rxn not in model_rxns:
+                model.remove_reaction(rxn)
+        reaction_id = 'tmp_objective'
+        reaction_ex = Reaction(Direction.Forward, {compound_id: -1})
+        model.database.set_reaction(reaction_id, reaction_ex)
+        model.add_reaction(reaction_id)
 
-            v = prob.namespace()
+        prob = solver.create_problem()
+
+        v = prob.namespace()
 
 
-            # We keep track of temporary constraints from the last optimization so
-            # that we can remove them during the next call to solve(). This is
-            # necessary because removing the constraint immediately after solving
-            # a model can invalidate the solution in some solvers.
-            temp_constr = []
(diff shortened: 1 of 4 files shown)


keitht547:
The documentation of the functions involved in this command need to be updated/completed before merging.

psamm/commands/completepath.py
```diff
@@ -0,0 +1,280 @@
...
+        # Add exchange and transport reactions to database
+        model_complete, weights = create_extended_sink_source_model(
+            self._model,
+            db_penalty=self._args.db_penalty,
+            source_penalty=self._args.source_penalty,
+            sink_penalty=self._args.sink_penalty,
+            tp_penalty=self._args.tp_penalty,
+            penalties=penalties)
+
+        implicit_sinks = not self._args.implicit_sinks
```
jonls: This should be `... = self._args_implicit_sinks`? I seems that the `not` accidentally reverses the logic.

psamm/commands/completepath.py
```diff
@@ -0,0 +1,280 @@
...
+        try:
+            added_reactions, no_bounds_reactions = gapfill(
+                model_complete, core, blocked, exclude, solver=solver,
+                epsilon=epsilon, v_max=v_max, weights=weights,
+                implicit_sinks=implicit_sinks,
+                allow_bounds_expansion=self._args.allow_bounds_expansion)
+        except GapFillError as e:
+            self._log_epsilon_and_fail(epsilon, e)
+
+        if self._args.gapfill is True:
```
jonls: `if self._args.gapfill:`

psamm/commands/completepath.py
```diff
@@ -0,0 +1,280 @@
...
+            '--implicit-sinks', action='store_true',
+            help='Do not include implicit sinks when gap-filling')
+        parser.add_argument(
+            '--allow-bounds-expansion', action='store_true',
+            help=('Allow GapFill to propose expansion of flux bounds. This'
+                  ' includes turning irreversible reactions reversible.'))
+        parser.add_argument(
+            '--gapfill', action='store_true',
+            help='Run the gapfill algorithm on the model with separated '
+                 'sink and source reactions in each compatment.')
```
jonls: What is `--gapfill` used for? Isn't it just the same as running the `gapfill` command?

psamm/commands/completepath.py
```diff
@@ -0,0 +1,280 @@
...
+    Uses an extended model and reactions identified through the gapfilling procedure
+    to create a new LP problem and maximize the production of a target compound. This
+    function will generate an LP problem with implicit sinks for all compounds
+    except for ones involved in the gap-filling reactions. An L1 minimized FBA problem
+    is then solved and the result is a flux based synthesis pathway for a specific compound.
+    This returns a reaction ID, flux value, and reaction entry for each reaction in the metabolic
+    model.
+
+    Args:
+        model: :class:`psamm.datasource.native.NativeModel`.
```
jonls: This seems to be a `MetabolicModel` not `NativeModel`.

psamm/commands/completepath.py
```diff
@@ -0,0 +1,280 @@
...
+        _z = z.set(model.reactions)
+        _v = v.set(model.reactions)
+
+        prob.add_linear_constraints(_z >= _v, _v >= -_z)
+
+        objective = z.expr(
+            (reaction_id, -1)
+            for reaction_id in model.reactions)
+        prob.set_objective(objective)
+        prob.solve(lp.ObjectiveSense.Maximize)
```
jonls: I would do a minimize here instead of maximizing the negative zs. If you minimize, you can also create the objective simply with `objective = z.sum(model.reactions)`.

psamm/commands/completepath.py
keitht547: This would be running the gapfill but with the modified extended model that includes separate sources/sinks in every model compartment. So the results will be slightly different than the other gapfill implementation.

psamm/commands/completepath.py
keitht547: You're right. the not was leftover from how the gapfill command worked with '--no-implicit-sinks'

PR #274, opened 2023-06-24
Change pd.append to pd.concat for pandas 2.0

zhenjiaofenjie:
Since pandas 2.0, pd.append is no longer supported, which leads to a broken manual_curation function. 
It can be easily fixed by migrating to pd.concat.

Greetings from Jing!


yingzhanguri:
We merged the pull request to develop

